PR #4, opened 2019-07-05
Move generated Python code under grakn_protocol package

vmax:

Previously, generated Python package contained `keyspace` and `session` packages — which have pretty generic names to be put into user's directory with installed Python packages. This PR moves them under `grakn_protocol` root package.

- Add `python_proto.py` tool which rewires imports in `.proto` files according to new structure
- Add `python-package-structure` which preprocesses `.proto` files so they have structure we want in generated Python code
- Replace `imports` in `py_library` so that correct folder is imported (will be needed after `@stackb_rules_proto` is upgraded to latest `master`)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -18,7 +18,7 @@
 
 package(default_visibility = ["//visibility:public"])
 
-exports_files(["VERSION"])
+exports_files(["VERSION", "python_proto.py"])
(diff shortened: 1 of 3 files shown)


vmax:
Superseded by #5

PR #27, opened 2020-04-03
Add prefilling of various response methods

adammitchelldev:

Currently, the protocol only returns IDs for every answer. However, it is very rare to only require the ID for a concept e.g. for an attribute you are most likely interested in its value. The goal of this PR is to mitigate the case where hundreds of concept IDs are returned for Attributes from a query, and the client program is going to do a sync RPC to get the value of each.

By taking a selection of example queries run against protocol, we can determine that there are some common requests that are always bounded (if the returned result is a concept, pre-filling it wouldn't lead to recursion). In the case that the returned concept would cause this kind of explosion, or for some reason the results can't be pre-fetched, the `isPrefilled` field notifies a client that this concept wasn't pre-filled (as opposed to just missing various pre-filled parts). Currently there is no way to explicitly enable pre-filling.

This is a static and simple variation on the "time travelling RPC" concept, where we are returning the results of another RPC call on the result of a first (see Cap'n Proto), although we are a long way from achieving this as a general pattern right now.

### DIFF
--- a/session/Concept.proto
+++ b/session/Concept.proto
@@ -107,24 +107,14 @@ message Method {
             SchemaConcept.SetLabel.Res schemaConcept_setLabel_res = 202;
             SchemaConcept.GetSup.Res schemaConcept_getSup_res = 203;
             SchemaConcept.SetSup.Res schemaConcept_setSup_res = 204;
-            SchemaConcept.Sups.Iter schemaConcept_sups_iter = 205;
-            SchemaConcept.Subs.Iter schemaConcept_subs_iter = 206;
 
             // Rule method responses
             Rule.When.Res rule_when_res = 300;
             Rule.Then.Res rule_then_res = 301;
 
-            // Role method responses
-            Role.Relations.Iter role_relations_iter = 401;
-            Role.Players.Iter role_players_iter = 402;
-
             // Type method responses
             Type.IsAbstract.Res type_isAbstract_res = 500;
             Type.SetAbstract.Res type_setAbstract_res = 501;
-            Type.Instances.Iter type_instances_iter = 502;
-            Type.Keys.Iter type_keys_iter = 503;
-            Type.Attributes.Iter type_attributes_iter = 504;
-            Type.Playing.Iter type_playing_iter = 505;
             Type.Has.Res type_has_res = 506;
             Type.Key.Res type_key_res = 507;
             Type.Plays.Res type_plays_res = 508;
@@ -137,7 +127,6 @@ message Method {
 
             // RelationType method responses
             RelationType.Create.Res relationType_create_res = 700;
-            RelationType.Roles.Iter relationType_roles_iter = 701;
             RelationType.Relates.Res relationType_relates_res = 702;
             RelationType.Unrelate.Res relationType_unrelate_res = 703;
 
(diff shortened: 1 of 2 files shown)


session/Concept.proto
```diff
@@ -650,6 +646,7 @@ message Relation {
         message Req {}
         message Iter {
             int32 id = 1;
```
haikalpribadi, line 648: Are we meant to keep this? It should be deleted like the rest, right?

session/Session.proto
```diff
@@ -98,18 +97,28 @@ message Transaction {
...
+
                 Query.Iter.Res query_iter_res = 2;
                 GetAttributes.Iter.Res getAttributes_iter_res = 3;
                 Method.Iter.Res conceptMethod_iter_res = 4;
             }
         }
+        message Batch {
+            message Options {
+                int32 batch_size = 1;
+            }
```
haikalpribadi, line 113: Are we adding this as a placeholder to add more options in the future?

session/Session.proto
```diff
@@ -98,18 +97,28 @@ message Transaction {
...
+            message Req {
+                int32 id = 1;
+                Options batch_options = 2;
+            }
```
haikalpribadi, line 117: Where are using `Iter.Batch.Req`?

session/Session.proto
```diff
@@ -98,18 +97,28 @@ message Transaction {
...
+            message Res {
+                int32 id = 1;
+            }
```
haikalpribadi, line 120: I like this!

session/Concept.proto
adammitchelldev: Ah yes, good catch.

session/Session.proto
adammitchelldev: Kind of, it's because we need any options in two places (in `Query.Req` and `Iter.Batch.Req`) so I wanted to make sure that's properly captured now.

session/Session.proto
adammitchelldev: It will be used by the client's inner iterator to request another batch of results for an existing iterator.

session/Session.proto
haikalpribadi: 👍

session/Session.proto
haikalpribadi: got it 👍

adammitchelldev:

Currently, the protocol only returns IDs for every answer. However, it is very rare to only require the ID for a concept e.g. for an attribute you are most likely interested in its value. The goal of this PR is to mitigate the case where hundreds of concept IDs are returned for Attributes from a query, and the client program is going to do a sync RPC to get the value of each.

By taking a selection of example queries run against protocol, we can determine that there are some common requests that are always bounded (if the returned result is a concept, pre-filling it wouldn't lead to recursion). In the case that the returned concept would cause this kind of explosion, or for some reason the results can't be pre-fetched, the `isPrefilled` field notifies a client that this concept wasn't pre-filled (as opposed to just missing various pre-filled parts). Currently there is no way to explicitly enable pre-filling.

This is a static and simple variation on the "time travelling RPC" concept, where we are returning the results of another RPC call on the result of a first (see Cap'n Proto), although we are a long way from achieving this as a general pattern right now.

PR #29, opened 2020-04-08
Refactor protocol for faster iterator requests

adammitchelldev:

The existing protocol made a round-trip to get an iterator ID and then individual round-trips for each result.

The goal of this PR is to optimise this design to require minimal roundtrips, reducing the effect of latency on queries with many results and the concept API.

* All iteration requests and responses are now separated from single-response methods, which makes their behaviour difference much clearer. `Req` and `Res`, and `Iter.Req` and `Iter.Res` are now all 1-to-1 mappings.
* Iterator requests no longer return an iterator ID upfront. Instead, they begin iterating immediately.
* Iterating no longer requires a single new request for each element. Iteration results are streamed as messages from the client to the server until the stream terminates (with a done response or...)
* Iterators can now temporarily terminate with a new `Iter.Res iteratorId` result, which gives the user an iterator ID that they can use to continue iterating with `Iter.Req iteratorId`.
* Users can specify the desired batching behaviour of any iteration request with `Iter.Req.Options`, primarily the batch size (optionally asking for all results).


session/Concept.proto
```diff
@@ -151,56 +122,85 @@ message Method {
...
-            Thing.Keys.Iter.Res thing_keys_iter_res = 902;
-            Thing.Attributes.Iter.Res thing_attributes_iter_res = 903;
-            Thing.Relations.Iter.Res thing_relations_iter_res = 904;
-            Thing.Roles.Iter.Res thing_roles_iter_res = 905;
-
-            // Relation iterator responses
-            Relation.RolePlayersMap.Iter.Res relation_rolePlayersMap_iter_res = 1000;
-            Relation.RolePlayers.Iter.Res relation_rolePlayers_iter_res = 1001;
+        message Req {
+            oneof iter_req {
```
haikalpribadi, line 139: let's rename `iter_req` to `req` to be consistent with other `oneof` variable names (being just the lowercasing of their owning `message`).

session/Concept.proto
```diff
@@ -151,56 +122,85 @@ message Method {
...
+
+                // Attribute iterator requests
+                Attribute.Owners.Iter.Req attribute_owners_iter_req = 1101;
+            }
+        }
 
-            // Attribute iterator responses
-            Attribute.Owners.Iter.Res attribute_owners_iter_res = 1101;
+        message Res {
+            oneof iter_res {
```
haikalpribadi, line 173: let's rename `iter_res` to `res` to be consistent with other `oneof` variable names (being just the lowercasing of their owning `message`).

session/Session.proto
```diff
@@ -64,11 +64,8 @@ message Transaction {
         oneof req {
             Open.Req open_req = 1;
             Commit.Req commit_req = 2;
-            Query.Req query_req = 3;
-            Iter.Req iterate_req = 4;
```
haikalpribadi, line 68: Shouldn't we keep `Iter.Req`? I see that `Iter.Res` is still returned (which I expect). The `stream()` can only be sending `Transaction.Req` and receiving `Transaction.Res`, right.

session/Concept.proto
adammitchelldev: The rationale here is that this is in an `Iter.Req` message, and all of the children of `Iter.Res` are named: `xxx_xxx_iter_req`, so I think this way is actually more consistent.

haikalpribadi: Right, I can see that. But in this case, we've been using just the name of parent on-level up, rather than two-level up. For example, `oneof answer`. So let's keep it just `req` to be consistent.

session/Session.proto
adammitchelldev: I actually intended to keep it separate from `Transaction.Req` but I've realised that this doesn't work because the stream is `Transaction.Req -> Transaction.Res`. I've put the `Req` back now.

session/Session.proto
```diff
@@ -64,11 +64,9 @@ message Transaction {
...
+            Iter.Req iter_req = 3;
```
haikalpribadi, line 67: Don't forget to set the ID number back to 4, @adammitchelldev

haikalpribadi:
Amazing! Well done, @adammitchelldev !

PR #42, opened 2020-07-28
Updating 2.0 concept protocol

adammitchelldev:

To align protocol with the 2.0 Concept API on the server side. A lot of this is simple renaming but some methods have changed slightly. The biggest changes were overridden attributes, roles owned by relations and removal of "key" in favour of parameters for "has".

- Renamed methods to be consistent with 2.0
- Added and removed methods where necessary

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -51,8 +51,8 @@ message Method {
 
             // RelationType method requests
             RelationType.Create.Req relationType_create_req = 700;
+            RelationType.Role.Req relationType_role_req = 701;
             RelationType.Relates.Req relationType_relates_req = 702;
-            RelationType.Unrelate.Req relationType_unrelate_req = 703;
 
             // AttributeType method requests
             AttributeType.Create.Req attributeType_put_req = 800;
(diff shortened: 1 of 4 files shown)


protobuf/concept.proto
```diff
@@ -335,7 +327,7 @@ message Role {
         message Iter {
```
haikalpribadi, line 327: `Role` should be renamed to `RoleType`

protobuf/transaction.proto
```diff
@@ -204,15 +180,15 @@ message Transaction {
 
     message ConceptMethod {
         message Req {
-            string id = 1;
+            bytes id = 1;
```
haikalpribadi: `iid`

protobuf/transaction.proto
```diff
@@ -204,15 +180,15 @@ message Transaction {
...
             Method.Req method = 2;
         }
         message Res {
             Method.Res response = 1;
         }
         message Iter {
             message Req {
-                string id = 1;
+                bytes id = 1;
```
haikalpribadi: `iid`

PR #44, opened 2020-08-07
Split Relation.Players into AllPlayers and PlayersForRoles

alexjpwalker:

To fix Relation.players() always returning an empty collection when called from any external Grakn client (e.g. client-java).

In https://github.com/graknlabs/grakn-2.0/pull/13, we updated the Grakn RPC to take in two different kinds of messages for retrieving roleplayers from a relation; one filtered by roles, and one that isn't.

Then we updated the protocol to represent the new message format, splitting Relation.Players into AllPlayers and PlayersForRoles.

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -613,13 +613,19 @@ message Relation {
         }
     }
 
-    message Players {
+    message AllPlayers {
+        message Iter {
+            message Req {}
+            message Res {
+                Concept thing = 1;
+            }
+        }
+    }
+
+    message PlayersForRoles {
         message Iter {
             message Req {
-                message All {}
-                message ForRoles {
-                    repeated Concept roles = 1;
-                }
+                repeated Concept roles = 1;
             }
             message Res {
                 Concept thing = 1;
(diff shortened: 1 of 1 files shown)


protobuf/concept.proto
```diff
@@ -615,7 +616,10 @@ message Relation {
     message Players {
         message Iter {
             message Req {
-                repeated Concept roles = 1;
+                message All {}
+                message ForRoles {
+                    repeated Concept roles = 1;
+                }
```
haikalpribadi, line 622: Every `message` inside this `message Relation` object represents a `method` on the `Relation` class in Grakn, @alexjpwalker. As you may notice `playersMap`, `relate`, and `unrelate` are all methods on the `Relation` class. Given that the goal of this PR is to replace `players` method with 2 different methods, we should be replacing the `message Players` object altogether with 2 new ones.

alexjpwalker: The 2 different methods will both be named `players`. What should we do in this case?

alexjpwalker: We decided to split the message up into AllPlayers  and PlayersForRoles.

PR #59, opened 2020-08-25
Align Type methods in protocol with core

alexjpwalker:

To align the Type methods in protocol with core.

- Add VALUE_TYPE.OBJECT
- Add support for role overrides
- Make SetRelates response empty
- Allow GetRelatesForRoleLabel to return null
- Pull IsAbstract up from ThingType to Type (but SetAbstract remains in ThingType)
- Add GetOwners to AttributeType
- Add UnsetRelates to RelationType

We also set up `protocol` for CI/CD using grabl.io.

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -301,6 +301,7 @@ message TypeMethod {
             RelationType.Create.Req relationType_create_req = 700;
             RelationType.GetRelatesForRoleLabel.Req relationType_getRelatesForRoleLabel_req = 701;
             RelationType.SetRelates.Req relationType_setRelates_req = 702;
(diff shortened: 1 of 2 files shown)


protobuf/concept.proto
```diff
@@ -637,17 +648,27 @@ message RelationType {
             string label = 1;
         }
         message Res {
-            Type role = 1;
+            oneof role {
```
alexjpwalker, line 651: This change ensures that GetRelatesForRoleType is allowed to return null.

alexjpwalker:
Do not merge yet - working on `automation.yml`

haikalpribadi (approved):

PR #63, opened 2020-09-03
Rename schema to encoding to match grakn 2.0

flyingsilverfin:
Following https://github.com/graknlabs/grakn-2.0/pull/35, we align terminology to not use `schema` but `encoding` for concept root type information.

Rename `schema` to `encoding` to match terminology in grakn 2.0

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -391,7 +391,7 @@ message TypeMethod {
 message Type {
     string label = 1;
     string scope = 2;
-    ENCODING ENCODING = 3;
+    ENCODING encoding = 3;
     AttributeType.VALUE_TYPE valueType = 4;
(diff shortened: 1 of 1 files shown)


protobuf/concept.proto
```diff
@@ -391,11 +391,11 @@ message TypeMethod {
 message Type {
     string label = 1;
     string scope = 2;
-    SCHEMA schema = 3;
+    ENCODING ENCODING = 3;
```
alexjpwalker, line 394: Rename to `encoding` (lowercase)

alexjpwalker (changes requested):
Addressed a typo in Type.Encoding

alexjpwalker (approved):

PR #64, opened 2020-09-10
Rules are not types and gRPC field compaction

flyingsilverfin:
We split rules from conceptually being a Type in the protocol. They are still viewed as part of the world of "Concepts" in grakn, but as separate elements of the schema, only having a label in common with types.

* Split `Rule` from being part of `Type`
* rearrange `concept.proto` to group the `Concept` definitions at the top, specialised messages below
* compact and clean up numbering ranges of messages - won't affect us (breaking backwards compat anyway)

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -32,76 +32,105 @@ message Concept {
     }
 }
 
+// Type
 
-// Thing
+message Type {
+    string label = 1;
+    string scope = 2;
+    ENCODING encoding = 3;
+    AttributeType.VALUE_TYPE valueType = 4;
+    bool root = 5;
 
-message ThingMethod {
-    message Req {
-        oneof req {
-            // Thing method requests
-            Thing.Delete.Req thing_delete_req = 900;
-            Thing.GetType.Req thing_getType_req = 901;
-            Thing.IsInferred.Req thing_isInferred_req = 902;
-            Thing.SetHas.Req thing_setHas_req = 906;
-            Thing.UnsetHas.Req thing_unsetHas_req = 907;
-
-            // Relation method requests
-            Relation.AddPlayer.Req relation_addPlayer_req = 1002;
-            Relation.RemovePlayer.Req relation_removePlayer_req = 1003;
-        }
-    }
-    message Res {
-        oneof res {
-            // Thing method responses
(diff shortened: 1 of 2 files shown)


alexjpwalker (approved):
These changes look good

protobuf/concept.proto
```diff
@@ -32,76 +32,105 @@ message Concept {
...
-                Relation.GetPlayers.Iter.Res relation_getPlayers_iter_res = 1000;
-                Relation.GetPlayersByRoleType.Iter.Res relation_getPlayersByRoleType_iter_res = 1001;
+                string pattern = 1;
+            }
+        }
+    }
 
-                // Attribute iterator responses
-                Attribute.GetOwners.Iter.Res attribute_getOwners_iter_res = 1101;
+    message Then {
```
haikalpribadi, line 122: Following our new method naming convention on Concept API, the methods on rules should be `getWhen()` and `getThen()`, which means the message names here needs to reflect that: `GetWhen`, and `GetThen`. Make sure to apply this onto the rest of the variable names through this proto file.

protobuf/transaction.proto
```diff
@@ -219,5 +219,15 @@ message Transaction {
...
         }
+
+        message Rule {
+            message Req {
+                string label = 1;
+                RuleMethod.Req method = 2;
+            }
+            message Res {
+                RuleMethod.Res response = 1;
+            }
```
haikalpribadi, line 230: These newly introduced messages are not used anywhere. I presume they need to be used in `Transaction.Req` and `Transaction.Res` above?

haikalpribadi (changes requested):

protobuf/concept.proto
haikalpribadi: And make sure the method names on the server and client side concept API are implemented accordingly too, @flyingsilverfin

flyingsilverfin: addressed

haikalpribadi: The message names here are still `When` and `Then` @flyingsilverfin . Look at the other message names. They're supposed to represent the method names, right? So these ones should be `GetWhen` and `GetThen`. Also, given the discussion during our call, you should also introduce `SetWhen` and `SetThen` now, right?

haikalpribadi (approved):

PR #66, opened 2020-09-14
Delete Rule.GetWhen and GetThen

alexjpwalker:

To delete redundant Rule methods

Delete Rule.GetWhen and GetThen


haikalpribadi (approved):
@alexjpwalker pointed out that we don't even need `GetWhen` and `GetThen` at all, @flyingsilverfin, because the `when` and `then` patterns are already stored in the `Rule` message!

PR #68, opened 2020-09-16
Create the new Query API

alexjpwalker:

To provide protocol messages that fit all the needs of our new Query API 2.0 - specifically Graql Match, Define, Undefine, Insert and Delete, while thinking about forwards compatibility.

Added QueryProto, a new protobuf file that contains two new messages: Query, and Graql.

Query is designed to mimic the structure of Concept; it contains Req, Res, Iter.Req and Iter.Res, which each come in pairs and choose from the various Graql methods available.

The Graql methods are encapsulated in a new message named Graql. Graql contains Match, Define, Undefine, Insert and Delete. These are the possible methods to choose from when constructing a Query.

This is a significant change from our previous structure, where all Graql queries had the same type, and were thus forced to also return the same type - which meant we had an awkward Answer object that could contain all sorts of things. Now, each query request has its own dedicated response type - for example, 'repeated Type' for Define, 'ConceptMap' for Match.Iter, and nothing for Delete.

This Answer object does still exist, purely because of AnswerGroup - whose structure is yet to be fleshed out - but it has been slimmed down substantially; only ConceptMap, AnswerGroup and Number remain as possible types of Answers. Number is not currently used, but it will be used in the future for Aggregates. We've also deleted Explanation while we determine how it will look in 2.0.

### DIFF
--- a/grpc/java/BUILD
+++ b/grpc/java/BUILD
@@ -30,6 +30,7 @@ java_grpc_compile(
         "//protobuf:session-proto",
         "//protobuf:transaction-proto",
         "//protobuf:answer-proto",
+        "//protobuf:query-proto",
         "//protobuf:concept-proto",
         "//protobuf:options-proto",
     ]
--- a/grpc/nodejs/BUILD
+++ b/grpc/nodejs/BUILD
@@ -25,6 +25,7 @@ node_grpc_compile(
         "//protobuf:database-proto",
         "//protobuf:session-proto",
         "//protobuf:transaction-proto",
+        "//protobuf:query-proto",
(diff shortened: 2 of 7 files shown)


haikalpribadi (approved):
All looks good in this PR, but 2 things come to mind for what we could do next:
1) Should we merge `answer.proto` into `query.proto` ?
2) Should we extract the messages in `TransactionProto` into `ConceptProto` ?

alexjpwalker:
The relevant messages from `TransactionProto` have now been merged into `ConceptProto`; the future of `answer.proto` is still up in the air.

PR #69, opened 2020-10-19
Defines return nothing, Rules have no setters

flyingsilverfin:
Rules can no longer set a `when` or `then`, and `define` queries no longer return anything.

* Remove `setWhen` and `setThen` messages in `RuleMethod`
* `Define` queries have no return data

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -218,16 +218,12 @@ message RuleMethod {
         oneof req {
             Rule.Delete.Req rule_delete_req = 100;
             Rule.SetLabel.Req rule_setLabel_req = 101;
-            Rule.SetWhen.Req rule_setWhen_req = 102;
-            Rule.SetThen.Req rule_setThen_req = 103;
         }
     }
     message Res {
         oneof res {
             Rule.Delete.Res rule_delete_res = 100;
             Rule.SetLabel.Res rule_setLabel_res = 101;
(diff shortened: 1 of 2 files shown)


protobuf/concept.proto
```diff
@@ -218,16 +218,12 @@ message RuleMethod {
...
-            Rule.SetThen.Req rule_setThen_req = 103;
         }
     }
     message Res {
         oneof res {
             Rule.Delete.Res rule_delete_res = 100;
             Rule.SetLabel.Res rule_setLabel_res = 101;
-            Rule.SetWhen.Res rule_setWhen_res = 102;
-            Rule.SetThen.Res rule_setThen_res = 103;
         }
```
alexjpwalker, line 227: Given that we recently decided to make rule definitions mutable via Graql Define, should we perhaps consider reintroducing these Concept methods?

alexjpwalker (changes requested):

protobuf/concept.proto
flyingsilverfin: as a rule, Rules should be `atomic` ie. either you modify the entire rule (when + then), delete the entire rule , or create an entire rule. This means that we either add a new `setBody` method (which would essentially just create and delete a rule), or just remove the functionality to modify a rule via concept-api. I think the latter is the simplest and closest aligned with the way we implement `putRule` in the server, where we delete and re-add the rule.

alexjpwalker (approved):

PR #70, opened 2020-10-23
Major simplification and flattening of all structures

alexjpwalker:

To make our protocol clean, simple and easy to read while maintaining type safety.

- Delete `Transaction.Iter.*`, `Query.Iter.*`, `ThingMethod.Iter.*` and `TypeMethod.Iter.*` and rename all `.Iter.Req`, `.Iter.Res` to just `.Req`, `.Res` while annotating the `.Res` messages with `@Iterable`. Add `Fetch.Req`, `done` and `continue` which are special message types used during iteration.
- Break up the `Grakn` service into `DatabaseService`, `SessionService`, `TransactionService` which all live in their respective files
- Merge all messages from `ThingMethod`, `TypeMethod` and `RuleMethod` into `Thing`, `Type`, `Rule`
- Introduce `ConceptManager` and move the methods that `ConceptManager` has in Core into this message
- Use snake_case everywhere (this is performed in accordance with the [Protocol Buffers Style Guide](https://developers.google.com/protocol-buffers/docs/style) and ensures that our compiled Java output always looks nice - i.e. no more switch cases named GETPLAYERSBYROLETYPE_RES, but rather proper snake casing)
- Introduce a request ID that should be unique into every `Transaction.Req` in order for the new Map-based structure to work on the client side. This allows concurrent requests to be processed out-of-order and still handled correctly.
- Allow the network latency to be sent with every request that supports iteration. It is used by the server's Iterator to correctly compensate for the network latency.


grpc/java/BUILD
```diff
@@ -46,6 +45,7 @@ java_library(
         "@maven//:io_grpc_grpc_protobuf",
         "@maven//:io_grpc_grpc_stub",
         "@maven//:io_grpc_grpc_api",
+        "@maven//:javax_annotation_javax_annotation_api",
```
alexjpwalker, line 48: This change makes our gRPC Java library compatible with Java 11 and Java 14.

protobuf/answer.proto
```diff
@@ -26,8 +26,8 @@ package grakn.protocol;
 
 message Answer {
     oneof answer {
-        AnswerGroup answerGroup = 1;
-        ConceptMap conceptMap = 2;
+        AnswerGroup answer_group = 1;
+        ConceptMap concept_map = 2;
```
alexjpwalker, line 30: In order for gRPC Java to assign sensible method names, we need to enforce a consistent casing convention. The preferred convention for protocol buffers is snake_case.

protobuf/concept.proto
```diff
@@ -22,212 +22,101 @@ option java_outer_classname = "ConceptProto";
...
-            // Relation method requests
-            Relation.AddPlayer.Req relation_addPlayer_req = 200;
-            Relation.RemovePlayer.Req relation_removePlayer_req = 201;
+            GetType.Req get_type_req = 1;
+            GetThing.Req get_thing_req = 2;
+            GetRule.Req get_rule_req = 3;
+            PutEntityType.Req put_entity_type_req = 4;
+            PutAttributeType.Req put_attribute_type_req = 5;
+            PutRelationType.Req put_relation_type_req = 6;
+            PutRule.Req put_rule_req = 7;
```
alexjpwalker, line 35: These `ConceptManager` methods from Core were originally in `Transaction` in protobuf. They are now instead grouped into a matching `ConceptManager.Req` message.

protobuf/concept.proto
```diff
@@ -22,212 +22,101 @@ option java_outer_classname = "ConceptProto";
...
-            RelationType.Create.Req relationType_create_req = 500;
-            RelationType.GetRelatesForRoleLabel.Req relationType_getRelatesForRoleLabel_req = 501;
-            RelationType.SetRelates.Req relationType_setRelates_req = 502;
-            RelationType.UnsetRelates.Req relationType_unsetRelates_req = 503;
-
-            // AttributeType method requests
-            AttributeType.Put.Req attributeType_put_req = 600;
-            AttributeType.Get.Req attributeType_get_req = 601;
-            AttributeType.GetRegex.Req attributeType_getRegex_req = 602;
-            AttributeType.SetRegex.Req attributeType_setRegex_req = 603;
```
alexjpwalker, line 126: Given that these TypeMethods are just messages that take in a Type, and there was already a message called Type, we decided to move them there for simplicity.

protobuf/concept.proto
```diff
@@ -286,72 +222,67 @@ message Thing {
...
-            message Res {
-                Thing attribute = 1;
-            }
+        message Req {
+            repeated Type attribute_types = 1;
+            bool keys_only = 2;
+        }
+        // @Iterable
+        message Res {
+            Thing attribute = 1;
```
alexjpwalker, line 231: Having an intermediate message name `Iter` seemed unnecessarily verbose. We now flag iterable messages using comments.

protobuf/database.proto
```diff
@@ -22,6 +22,15 @@ option java_outer_classname = "DatabaseProto";
...
 
+service DatabaseService {
+
+    rpc contains (Database.Contains.Req) returns (Database.Contains.Res);
+    rpc create (Database.Create.Req) returns (Database.Create.Res);
+    rpc all (Database.All.Req) returns (Database.All.Res);
+    rpc delete (Database.Delete.Req) returns (Database.Delete.Res);
+
+}
+
```
alexjpwalker, line 33: Originally all "Service" methods lived in `grakn.proto` - we split them out into three services that each have a clearer purpose.

protobuf/transaction.proto
```diff
@@ -26,68 +26,46 @@ import "protobuf/query.proto";
...
-            GetRule.Req getRule_req = 8;
-            PutEntityType.Req putEntityType_req = 9;
-            PutAttributeType.Req putAttributeType_req = 10;
-            PutRelationType.Req putRelationType_req = 11;
-            PutRule.Req putRule_req = 12;
-            ThingMethod.Req thingMethod_req = 13;
-            TypeMethod.Req typeMethod_req = 14;
-            RuleMethod.Req ruleMethod_req = 15;
-            //Explanation.Req explanation_req = 16;
+            Fetch.Req fetch_req = 4;
```
alexjpwalker, line 44: `Fetch.Req` is an indirect replacement for the old `Iter.Req` - it must be called in combination with a request ID that is already being processed on the server, and is used to fetch the next batch of answers.

protobuf/transaction.proto
```diff
@@ -26,68 +26,46 @@ import "protobuf/query.proto";
...
-        message Res {
-            oneof res {
-                bool done = 1;
-                int32 iteratorID = 2;
-                Query.Iter.Res query_iter_res = 3;
-                ThingMethod.Iter.Res conceptMethod_thing_iter_res = 4;
-                TypeMethod.Iter.Res conceptMethod_type_iter_res = 5;
-            }
+            bool done = 2;
+            bool continue = 3;
```
alexjpwalker, line 60: The boolean flags `done` and `continue` are special response types that only appear during iteration.

If a response contains `done`, that means iteration is done and there are no more answers. If a response contains `continue`, that signals that the client should issue a `Fetch.Req` if it wants to fetch more answers.

protobuf/transaction.proto
```diff
@@ -102,7 +80,9 @@ message Transaction {
             Type type = 2;
             Options options = 3;
         }
-        message Res {}
+        message Res {
+            int32 processingTimeMillis = 1;
```
alexjpwalker, line 84: When a client is measuring network latency, they should subtract this value in their calculation. This tells the client how long the server spent doing the work to open the Transaction.

protobuf/transaction.proto
```diff
@@ -26,68 +26,46 @@ import "protobuf/query.proto";
...
+    rpc stream (stream Transaction.Req) returns (stream Transaction.Res);
+}
+
 message Transaction {
 
     message Req {
-        map<string, string> metadata = 1000;
+        string id = 1;
+        map<string, string> metadata = 2;
+        int32 latencyMillis = 3;
```
alexjpwalker, line 42: Each request must now be uniquely identified by ID in order for the new Map-based structure to work on the client side. This allows concurrent requests to be processed out-of-order and still handled correctly.

The network latency should be sent with every request that supports iteration. It is used by the server's Iterator to correctly compensate for the network latency.

protobuf/transaction.proto
haikalpribadi: This should just be a `bool continue`

haikalpribadi (approved):

PR #71, opened 2020-11-03
Add checkstyle, add release workflow in Grabl, delete CircleCI file

alexjpwalker:

To allow protocol to build in Grabl

Add checkstyle, add release workflow in Grabl, delete CircleCI file, and ensure that every file has an AGPL license


.bazelrc
```diff
@@ -1,3 +1,20 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
 build --incompatible_strict_action_env
 run --incompatible_strict_action_env
 test --incompatible_strict_action_env
```
lolski, line 20: The bazelrc doesn't have the buildbuddy stuff here. is it expected?

Compare it with the bazelrc of grakn-core for example: https://github.com/graknlabs/grakn-2.0/blob/master/.bazelrc

lolski (commented):
Looks mostly good, but I've got one question.

.bazelrc
alexjpwalker: What's missing? It looks good to me

lolski: ah lol yeah it's actually there when i expanded it. was just delusional

lolski (approved):

PR #76, opened 2020-11-24
Allow multiple answers batched within one grpc message 

lriuui0x0:

We have observed performance hit caused by sending large amount of grpc calls, to overcome the issue, we allow multiple answers to be batched within one grpc message and we batch them in one grpc call. This way we reduce the number calls we need to make to grpc.

- Make the messages that server may stream back multiple answers into `repeated` field.

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -226,7 +226,6 @@ message Thing {
             repeated Type attribute_types = 1;
             bool keys_only = 2;
         }
-        // @Iterable
         message Res {
             repeated Thing attribute = 1;
         }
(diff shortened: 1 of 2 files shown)


protobuf/concept.proto
```diff
@@ -228,15 +228,15 @@ message Thing {
         }
         // @Iterable
```
alexjpwalker, line 229: Do we still need `@Iterable` in all of these places?

Given that the thing in the `Res` is now `repeated`, I'd say it's now explicit enough that these things are iterable, so we can delete these comments.

alexjpwalker (changes requested):
Left just one comment

protobuf/concept.proto
lriuui0x0: Fixed.

alexjpwalker (approved):

PR #78, opened 2020-11-27
Typescript compilation

Phillammon:

Compile protobuf files to typescript rather than raw javascript in the nodejs target.

Create genrule calling protoc compiler directly with the appropriate plugins to output both javascript and typescript type definition files when built.


.gitignore
```diff
@@ -66,3 +66,6 @@ logs/
 
 # Python cache #
 __pycache__
+
+# Node Modules #
+node_modules
```
Phillammon, line 71: Having added node_modules, we do *not* want to add it to git

WORKSPACE
```diff
@@ -15,7 +15,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-workspace(name = "graknlabs_protocol")
+workspace(
+    name = "graknlabs_protocol",
+    managed_directories = {"@npm": ["node_modules"]},
```
Phillammon, line 20: As we've added node modules, we want to be able to use the modules as targets, which necessitates managing it as a directory.

WORKSPACE
```diff
@@ -109,4 +112,23 @@ maven(artifacts)
...
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/2.2.2/rules_nodejs-2.2.2.tar.gz"],
+)
+
+load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "npm_install")
+node_repositories()
+npm_install(
+  name = "npm",
+  package_json = "//:package.json",
+  package_lock_json = "//:package-lock.json",
+)
```
Phillammon, line 134: Performing an npm_install is required in order to get said node modules up to date and in the build, and we need to set up the nodejs rules so that we can fetch them in. (The node modules specifically required are the protoc compiler and the gen_grpc plugin for it)

grpc/nodejs/BUILD
```diff
@@ -16,21 +16,50 @@
...
+        "protobuf/grakn_grpc_pb.js",
+        "protobuf/grakn_grpc_pb.d.ts",
+        "protobuf/options_pb.d.ts",
+        "protobuf/options_pb.js",
+        "protobuf/query_pb.d.ts",
+        "protobuf/query_pb.js",
+        "protobuf/session_pb.d.ts",
+        "protobuf/session_pb.js",
+        "protobuf/transaction_pb.d.ts",
+        "protobuf/transaction_pb.js",
```
Phillammon, line 42: Set of outputs changed to reflect we a) now only have a single service and b) we now have both a .js and a .d.ts file for every incoming proto file.

grpc/nodejs/BUILD
```diff
@@ -16,21 +16,50 @@
...
+    ],
+    cmd = "ln -s $(execpath @npm//grpc_tools_node_protoc_ts/bin:protoc-gen-ts).runfiles/npm/node_modules ./node_modules ;\
+        $(execpath @npm//grpc-tools/bin:grpc_tools_node_protoc) \
+            --plugin='protoc-gen-ts=$(rootpath @npm//:node_modules/grpc_tools_node_protoc_ts/bin/protoc-gen-ts)' \
+            --js_out='import_style=es6,binary:./$(@D)/' \
+            --grpc_out='./$(@D)/' \
+            --ts_out='./$(@D)/' \
+            --proto_path=`dirname $(execpath //:WORKSPACE)` \
+            $(execpaths //protobuf:proto-raw-buffers);",
```
Phillammon, line 51: This command is a bit brave, but divided by line:
- Symlinks the node runfiles required by the typescript plugin into the directory where the protocol compiler can see it
- Runs the node protoc compiler
- Using the protoc-gen-ts plugin to generate typescript type definitions in addition to the javascript definitions
- Outputting the javascript, typescript, and service definitions to the output directory specified by bazel ($(@D))
- Looking in the workspace for the relative paths to the .proto files
- using the .proto files as specified in the proto-raw-buffers filegroup as the inputs

grpc/nodejs/BUILD
```diff
@@ -16,21 +16,50 @@
...
+        tools = [
+            "@npm//grpc-tools/bin:grpc_tools_node_protoc",
+            "@npm//grpc_tools_node_protoc_ts/bin:protoc-gen-ts",
+            "@npm//:node_modules/grpc_tools_node_protoc_ts/bin/protoc-gen-ts",
+            "@npm//grpc_tools_node_protoc_ts",
+            "@npm//google-protobuf",
+            "//protobuf:proto-raw-buffers",
+            "//:WORKSPACE"
+        ],
```
Phillammon, line 60: The required tools for the above command. Workspace only has to be included in order to find the relative path for the .proto files.

grpc/nodejs/protocol.json
```diff
@@ -0,0 +1,69 @@
...
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+
+    /* Advanced Options */
+    "skipLibCheck": true,                     /* Skip type checking of declaration files. */
+    "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
+  }
+}
```
Phillammon, line 69: Standard tsconfig file generated automatically by tsc, for tsc.

package.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "protocol",
+  "version": "1.0.0",
+  "dependencies": {},
+  "devDependencies": {
+    "@bazel/typescript": "2.0.3",
+    "typescript": "3.7.5",
+    "grpc-tools": "^1.9.1",
+    "grpc_tools_node_protoc_ts": "^5.0.1"
+  }
+}
```
Phillammon, line 11: Node dependencies in order to compile typescript

protobuf/BUILD
```diff
@@ -76,6 +76,20 @@ proto_library(
...
+        "answer.proto",
+        "concept.proto",
+        "database.proto",
+        "grakn.proto",
+        "options.proto",
+        "query.proto",
+        "session.proto",
+        "transaction.proto",
+    ]
+)
```
Phillammon, line 91: The libraries we currently have pre-compile the .proto files, which is inconvenient as we wish to recompile them. Where we could (and in future should) build a full blown bazel rule extracting .src_file from the proto_library targets, for the time being providing a filegroup is equally legitimate

grpc/python/BUILD
```diff
@@ -36,14 +36,14 @@ python_grpc_compile(
...
-    src = ":protocol_src",
+    name = "protocol-pkg",
+    src = ":protocol-src",
     package = "grakn_protocol"
 )
 
 py_library(
     name = "protocol",
-    srcs = ["protocol_pkg"],
+    srcs = ["protocol-pkg"],
```
Phillammon, line 46: Renamed for consistency on the way past

grpc/nodejs/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/graknlabs/protocol"
+  },
+  "dependencies": {
+    "@grpc/grpc-js": "1.2.1"
+  },
+  "devDependencies": {
+    "typescript": "3.9.7"
+  }
```
Phillammon, line 17: Package.json for compiling the npm package

grpc/nodejs/BUILD
```diff
@@ -16,21 +16,72 @@
...
+    name = "assemble-npm",
+    target = ":protocol-package",
+)
+
+deploy_npm(
+    name = "deploy-npm",
+    target = ":assemble-npm",
+    snapshot = deployment["npm.snapshot"],
+    release = deployment["npm.release"],
 )
```
Phillammon, line 85: Compile, assemble, package and deploy the npm package, using the same mechanisms as client-nodejs

.grabl/automation.yml
```diff
@@ -42,6 +42,21 @@ build:
...
+        script: |
+          curl https://cli-assets.heroku.com/apt/release.key | sudo apt-key add -
+          wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+          sudo apt update -y
+          sudo apt install -y expect
+          export DEPLOY_NPM_USERNAME=$REPO_GRAKN_USERNAME
+          export DEPLOY_NPM_PASSWORD=$REPO_GRAKN_PASSWORD
+          export DEPLOY_NPM_EMAIL=$REPO_GRAKN_EMAIL
+          bazel run --define version=$(git rev-parse HEAD) //grpc/nodejs:deploy-npm -- snapshot
+        dependencies: [build, build-dependency, test]
```
Phillammon, line 59: Automate the process of deploying snapshots

Phillammon (commented):
Added final comments now that we have this in a functional state for node client 2.0

.gitignore
alexjpwalker, line 71: Remember to end each file with a newline - this is for compatibility with unix features
```suggestion
node_modules

```

WORKSPACE
alexjpwalker: I feel like we should rename `@npm` to `@node_modules` for clarity - what do you think?

We would make the same change in `client-nodejs`.

grpc/nodejs/BUILD
alexjpwalker: This is excellent documentation, nice work! I think we should add it to the code itself, in comments, so this doesn't get lost in the ether.

grpc/nodejs/BUILD
alexjpwalker: Again, let's add this as a comment in the code itself 🙂 Something like: "We use //:WORKSPACE to find the relative path for the .proto files".

.grabl/automation.yml
lolski: We don't have a job `test` do we? In that case this should be removed

grpc/nodejs/package.json
```diff
@@ -0,0 +1,18 @@
...
+}
```
alexjpwalker, line 18:
```suggestion
}

```

package.json
alexjpwalker: Do we need both `package.json` files?

WORKSPACE
```diff
@@ -109,4 +112,25 @@ maven(artifacts)
...
-)
\ No newline at end of file
+)
+
+#########################
+# Load NPM dependencies #
+#########################
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
```
lolski, line 122: Can we update the version of `rules_nodejs` in `@graknlabs_dependencies` with this one and then have the `WORKSPACE` file load from there instead? You can get @alexjpwalker 's help for the know-how.

protobuf/BUILD
alexjpwalker: Totally with you on this one - let's raise a GitHub issue for it (in `graknlabs/protocol`) and possibly put a `TODO` in the source code linking to the issue so it doesn't get lost.

alexjpwalker (changes requested):
Overall this looks excellent. I've highlighted a couple of trailing newlines and suggested that we add relevant documentation to GitHub and the source code where applicable.

package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "protocol",
+  "version": "1.0.0",
```
lolski, line 3: VERSION should match the version file shouldn't it? In that case, 2.0.0-alpha?

grpc/python/BUILD
lolski: Nice effort!

However, will a Bazel target with `-` rather than `_` work in Python @vmax ?

dependencies/graknlabs/repositories.bzl
```diff
@@ -21,5 +21,5 @@ def graknlabs_dependencies():
     git_repository(
         name = "graknlabs_dependencies",
         remote = "https://github.com/Phillammon/dependencies",
-        commit = "8526f49ce77fcd8e9029ab4b08d0fd5d0b4497f8", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
+        commit = "63bfceddc8493e743f54ab230239cd42eb1bccdf", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
```
alexjpwalker, line 24: Update this before merging

alexjpwalker (approved):
Approving - just remember to bump dependencies and verify that it builds in Grabl

lolski (approved):

PR #86, opened 2020-12-07
Extract rule behaviours to logic.proto from concept.proto

flyingsilverfin:
Rules belong to a class of non-"concept" structures in Grakn and the protocol, that we call `logic`s. This is reflected by introducing a `LogicManager`, and extracting all logic and rule RPC behaviours into a `logic.proto` definition file.

* Introduce `logic.proto`
* Extract any rule behaviours into `logic.proto`


protobuf/logic.proto
```diff
@@ -0,0 +1,96 @@
...
+
+message Rule {
+    string label = 1;
+    string when = 2;
+    string then = 3;
+
+    message Req {
+        string label = 1;
+        oneof req {
+            Rule.Delete.Req rule_delete_req = 100;
```
flyingsilverfin, line 73: shall i collapse these down to use small numbers now @alexjpwalker ?

alexjpwalker: I think 100 is fine- it adds what, 6 bits to the message size? That won't be noticeable

alexjpwalker (approved):
Seems good to me

PR #89, opened 2020-12-10
Consistent casing for enums + new npm package name

alexjpwalker:

Enforce consistent casing behaviour for enums, and give our npm package the right name.

- All enums are now PascalCased, as in the Protocol Buffers guide.
- The npm package output by `grpc/nodejs:BUILD` is now named `grakn-protocol`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "graknlabs-protocol",
+  "name": "grakn-protocol",
   "version": "2.0.0-alpha",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "graknlabs-grpc-protocol",
+  "name": "graknlabs-protocol",
```
flyingsilverfin, line 2: shouldn't this be "grakn-protocol" tho

alexjpwalker: It doesn't actually need to be - but it does make sense to keep consistent, so I've updated it to `grakn-protocol`.

flyingsilverfin (approved):

PR #90, opened 2020-12-10
Make grpc/python package correctly importable with pip

alexjpwalker:

Make grpc/python package correctly importable with pip.

- Do some magic that makes the `py_library` compatible with pip


alexjpwalker:
This fix was originally written by Max, but I reverted it because of a mistake where the tests were not importing the correct package. Now we reintroduce it.

lolski (approved):

PR #93, opened 2020-12-15
Simplify NodeJS proto compilation

vmax:

Simplify the `genrule` that builds our protocol for Typescript

* Move `//protobuf:proto-raw-buffers` into `srcs` since this is what they are
* Remove symlinking `node_modules` - not needed
* `dirname $(execpath //:WORKSPACE)` always evaluates to `.` (since `WORKSPACE` is a top-level file)

### DIFF
--- a/grpc/nodejs/BUILD
+++ b/grpc/nodejs/BUILD
@@ -71,24 +71,22 @@ genrule(
     #    Output typescript to (you guessed it) the genrule output directory
     #    Set the .proto file relative path to the WORKSPACE this command is being run from
     #    Use the .proto files found in the :proto-raw-buffers filegroup as inputs.
-    cmd = "ln -s $(execpath //grpc/nodejs:protoc-gen-ts).runfiles/npm/node_modules ./node_modules ;\
-        $(execpath //grpc/nodejs:grpc_tools_node_protoc) \
+    cmd = "$(execpath //grpc/nodejs:grpc_tools_node_protoc) \
             --plugin='protoc-gen-ts=$(rootpath @npm//:node_modules/grpc_tools_node_protoc_ts/bin/protoc-gen-ts)' \
             --js_out='import_style=commonjs,binary:./$(@D)/' \
             --grpc_out='grpc_js:./$(@D)/' \
             --ts_out='grpc_js:./$(@D)/' \
-            --proto_path=`dirname $(execpath //:WORKSPACE)` \
-            $(execpaths //protobuf:proto-raw-buffers);",
-    # Dependencies for the above command. the //:WORKSPACE label is required to locate the relative path for the .proto files so they can import each other.
+            --proto_path=. \
+            $(locations //protobuf:proto-raw-buffers);",
     tools = [
         "//grpc/nodejs:grpc_tools_node_protoc",
-        "//grpc/nodejs:protoc-gen-ts",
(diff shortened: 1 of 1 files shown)


grpc/nodejs/BUILD
```diff
@@ -71,24 +71,22 @@ genrule(
     #    Output typescript to (you guessed it) the genrule output directory
     #    Set the .proto file relative path to the WORKSPACE this command is being run from
     #    Use the .proto files found in the :proto-raw-buffers filegroup as inputs.
```
Phillammon, line 73: Comment needs updating if the functionality is changing

grpc/nodejs/BUILD
```diff
@@ -71,24 +71,22 @@ genrule(
...
-    cmd = "ln -s $(execpath //grpc/nodejs:protoc-gen-ts).runfiles/npm/node_modules ./node_modules ;\
-        $(execpath //grpc/nodejs:grpc_tools_node_protoc) \
+    cmd = "$(execpath //grpc/nodejs:grpc_tools_node_protoc) \
             --plugin='protoc-gen-ts=$(rootpath @npm//:node_modules/grpc_tools_node_protoc_ts/bin/protoc-gen-ts)' \
             --js_out='import_style=commonjs,binary:./$(@D)/' \
             --grpc_out='grpc_js:./$(@D)/' \
             --ts_out='grpc_js:./$(@D)/' \
-            --proto_path=`dirname $(execpath //:WORKSPACE)` \
```
Phillammon, line 80: `execpath //:WORKSPACE` is only . if this is being built from the top level workspace- if it's being brought in as an external dependency, this doesn't necessarily hold- can you confirm this doesn't break when pulled in externally?

grpc/nodejs/BUILD
```diff
@@ -71,24 +71,22 @@ genrule(
...
-            $(execpaths //protobuf:proto-raw-buffers);",
-    # Dependencies for the above command. the //:WORKSPACE label is required to locate the relative path for the .proto files so they can import each other.
+            --proto_path=. \
+            $(locations //protobuf:proto-raw-buffers);",
```
Phillammon, line 80: locations is recommended against in the documentation- as it resolves to either rootpaths or execpaths, we should probably explicitly say either rootpaths or execpaths as appropriate.

Phillammon (changes requested):

grpc/nodejs/BUILD
vmax: @Phillammon confirmed: i tried pulling `protocol` from `client-java` & `client-nodejs`, none of them broke

vmax: Also: `//grpc/nodejs:protocol` is not designed to be used as a Bazel dependency - it should only be pulled from NPM

grpc/nodejs/BUILD
vmax: @Phillammon addressed in https://github.com/graknlabs/protocol/pull/93/commits/97ce438beac7193c10fd7248ef76820ed1b9a59f

grpc/nodejs/BUILD
vmax: @Phillammon addressed in https://github.com/graknlabs/protocol/pull/93/commits/97ce438beac7193c10fd7248ef76820ed1b9a59f

Phillammon (approved):

PR #95, opened 2020-12-16
Use ts_grpc_compile macro to compile NodeJS TS/JS proto files

vmax:

Currently, the developer experience of compiling `@graknlabs_protocol` for NodeJS is suboptimal: we have to maintain a list of raw proto files in `//protocol:proto-raw-buffers` and it's generally inconsistent with how we compile protos in other languages (Python/Java)

Implement `ts_grpc_compile` macro which `//grpc/nodejs:protocol` will now utilize


grpc/nodejs/BUILD
```diff
@@ -37,54 +38,26 @@ nodejs_binary(
   templated_args = ["--nobazel_node_patches"],
 )
 
+py_binary(
```
lolski, line 41: what is this for @vmax ?

lolski: how is this related to the declaration to `ts_grpc_compile` below?

vmax: @lolski so the way we compile proto files is:

1) turn proto targets into their respective sources
2) use old genrule to generate js/ts files into one .tar.gz archive
3) use another rule which unpacks that archive into the files that we expect

(3) is covered by this `py_binary` target - it takes in an archive with generated sources and all of the filenames that are expected in the output and extracts them

Phillammon: Is it not possible to extract the source files via the proto_library .src property? This came up in research, but I was unclear on the feasibility.

vmax: @Phillammon rules and macros don't have access to target properties directly - only if targets choose to expose them via _providers_ (such as `ProtoInfo` in this case)

grpc/nodejs/unpack_proto_archive.py
```diff
@@ -0,0 +1,35 @@
...
+import tarfile
+
+_, arc, *files = sys.argv
+
+files_map = {}
+for f in files:
+    files_map[os.path.basename(f)] = f
+
+
+with tarfile.open(arc, 'r') as tar:
```
lolski, line 30: why do we write it in Python @vmax? can we just use `unzip` in `bash`, if it's a one-liner?

lolski (approved):

Phillammon (approved):

PR #97, opened 2020-12-22
Update how we pass requirements to assemble_pip

vmax:

Recently, we've upgraded how Python package dependencies are passed to `assemble_pip` rule. Therefore, all usages of `assemble_pip` need to be updated.

Pass `requirements.txt` to `assemble_pip`


grpc/python/BUILD
```diff
@@ -78,7 +80,7 @@ assemble_pip(
     author = "Grakn Labs",
     author_email = "grabl@grakn.ai",
     license = "AGPLv3+",
-    install_requires = [],
+    install_requires_file = "//grpc/python:requirements.txt",
```
lolski, line 83: shouldn't this be `requirements_file` @vmax ? I remember us having a discussion about that

vmax:
Looks like this change was already applied in https://github.com/graknlabs/protocol/commit/a8c3388d0f684bd3918a798e844da090fab39b35

PR #100, opened 2021-01-04
Support match, match aggregate, match group, and match group aggregate

lolski:

We have updated protocol to support not only `match` queries, but also `match aggregate`, `match group`, and `match group aggregate`.

- Extend the protocol by adding new message types: `Match.Aggregate`, `Match.Group`, and `Match.Group.Aggregate`
- Remove the ability to nest `AnswerGroup` which is no longer needed given the new answer structure.


protobuf/query.proto
```diff
@@ -58,6 +58,36 @@ message Graql {
...
+            }
+        }
+
+        message Group {
+            message Req {
+                string query = 1;
+            }
+
+            message Res {
+                repeated AnswerGroup answers = 1;
```
lolski, line 78: Should we be even more type-safe and make `AnswerGroup` more specific: `AnswerGroup` and `AnswerGroup.Aggregate`?

protobuf/answer.proto
```diff
@@ -26,9 +26,8 @@ package grakn.protocol;
 
 message Answer {
     oneof answer {
-        AnswerGroup answer_group = 1;
-        ConceptMap concept_map = 2;
-        Number number = 3;
+        ConceptMap concept_map = 1;
+        Number number = 2;
```
alexjpwalker, line 30: I wonder if we should pick a less ambiguous term here - `NumericAnswer numeric = 2`, perhaps?

protobuf/query.proto
```diff
@@ -58,6 +58,36 @@ message Graql {
...
             repeated ConceptMap answers = 1;
         }
+
+        message Aggregate {
+            message Req {
+                string query = 1;
+            }
+
+            message Res {
+                Number answers = 1;
```
alexjpwalker, line 68:
```suggestion
                Number answer = 1;
```

protobuf/query.proto
alexjpwalker, line 78:
```suggestion
                repeated AnswerGroup answer_groups = 1;
```

protobuf/query.proto
```diff
@@ -58,6 +58,36 @@ message Graql {
...
+                repeated AnswerGroup answers = 1;
+            }
+
+            message Aggregate {
+                message Req {
+                    string query = 1;
+                }
+
+                message Res {
+                    repeated AnswerGroup answers = 1;
```
alexjpwalker, line 87:
```suggestion
                    repeated AnswerGroup answer_groups = 1;
```

protobuf/query.proto
alexjpwalker: I think that makes sense, yes - it also follows the structure of the Java Graql repo. `GraqlMatch`, `GraqlMatch.Group`, `GraqlMatch.Aggregate`, `GraqlMatch.Group.Aggregate`

alexjpwalker (changes requested):
Made some suggestions.

protobuf/answer.proto
alexjpwalker: We decided not to.

alexjpwalker:
We decided to:

- move `Match`, `Insert`, `Delete` etc. inside `Query` (instead of where it currently lives, inside `Graql`) and delete `Graql`.
- delete `Answer`
- rename `AnswerGroup` to `ConceptMapGroup`
- create `NumberGroup`
- name the new query type messages: `MatchGroup`, `MatchAggregate`, `MatchGroupAggregate`

@lolski

protobuf/answer.proto
```diff
@@ -24,25 +24,22 @@ import "protobuf/concept.proto";
...
-}
-
 message ConceptMap {
     map<string, Concept> map = 1;
     string pattern = 2;
 }
 
+message ConceptMapGroup {
+    Concept owner = 1;
+    repeated ConceptMap concept_map = 2;
```
alexjpwalker, line 34: Let's make sure all our `repeated` fields have plural names- this is particularly beneficial to client Python, which uses the field names themselves as property names:
```py
answers: list = match_res.answers
```
is more readable than
```py
answers: list = match_res.answer
```
```suggestion
    repeated ConceptMap concept_maps = 2;
```

protobuf/answer.proto
```diff
@@ -24,25 +24,22 @@ import "protobuf/concept.proto";
...
+}
+
 message Number {
     string value = 1;
     // We use a string to contain the full representation of a number
 }
+
+message NumberGroup {
+    Concept owner = 1;
+    repeated Number number = 2;
```
alexjpwalker, line 44: See comment above

protobuf/query.proto
```diff
@@ -30,26 +30,23 @@ message Query {
...
-            Graql.Delete.Req delete_req = 100;
-            Graql.Define.Req define_req = 101;
-            Graql.Undefine.Req undefine_req = 102;
-            Graql.Match.Req match_req = 103;
-            Graql.Insert.Req insert_req = 104;
+            Delete.Req delete_req = 100;
+            Define.Req define_req = 101;
+            Undefine.Req undefine_req = 102;
+            Match.Req match_req = 103;
+            Insert.Req insert_req = 104;
```
alexjpwalker, line 37: We additionally need:
```protobuf
MatchAggregate.Req match_aggregate_req = 105;
MatchGroup.Req match_group_req = 106;
MatchGroupAggregate.Req match_group_aggregate_req = 107;
```
And the matching responses should be added to `Res`, below

alexjpwalker (changes requested):

protobuf/answer.proto
lolski: Fixed in https://github.com/graknlabs/protocol/pull/100/commits/8e16eed30e9a1221201d3b4ff85108dd6874ea86

protobuf/answer.proto
lolski: Fixed in https://github.com/graknlabs/protocol/pull/100/commits/8e16eed30e9a1221201d3b4ff85108dd6874ea86

protobuf/query.proto
lolski: Fixed in https://github.com/graknlabs/protocol/pull/100/commits/e6354fe87b04c8f37c5c81c8387d24c5fdcedfa4

alexjpwalker (approved):

PR #102, opened 2021-01-05
Refine the implementation of NumericGroup and Numeric

lolski:

We have refined the protocol for returning `Numeric` and `NumericGroup` which are used by match aggregate and match group aggregate respectively.

1. Enrich the representation of `Numeric` such that it can represent long, double, or NaN (not a number)
2. Fix the fact that `NumericGroup` has a list of `Number` rather than `Number`


protobuf/answer.proto
```diff
@@ -34,12 +34,17 @@ message ConceptMapGroup {
...
 }
 
-message Number {
-    string value = 1;
-    // We use a string to contain the full representation of a number
+message Numeric {
+    oneof value {
+        int64 long_value = 1;
+        double double_value = 2;
+        NaN nan_value = 3;
```
alexjpwalker, line 41: I think we can just use a `bool` here rather than defining a separate message - like we do elsewhere in protocol (for example `done` and `continue` in `Transaction.Res`)
```suggestion
        bool nan = 3;
```

lolski: Fixed in https://github.com/graknlabs/protocol/pull/102/commits/d859686566b66e3099c435bee09d8d48f0f67df0

alexjpwalker (approved):

PR #107, opened 2021-01-18
Add Cluster RPC definition which supports discovery protocol

lolski:

We have added Cluster RPC definition which supports discovery protocol. The definition would allow implementations such as Client Java and Grakn Cluster to discover and exchange cluster information.

- Adds Cluster RPC which supports discovery protocol. It exposes cluster information allowing for efficient communication between client and Grakn Cluster:
    - allows for discovering servers which belong to the cluster
    - allows for discovering the replicas belonging of a database


grpc/java/BUILD
```diff
@@ -38,6 +38,14 @@ java_grpc_compile(
...
 )
 
+java_grpc_compile(
+    name = "protocol-cluster-src",
+    deps = [
+        "//protobuf/cluster:cluster-proto",
+        "//protobuf/cluster:discovery-proto"
+    ]
+)
+
```
alexjpwalker, line 48: We could, maybe, split up the folders differently, and create `protobuf/core:answer-proto` (etc.) and `protobuf/cluster:cluster-proto` - what do you think?

I actually kind of prefer it the way it is now, but just throwing out an idea

protobuf/cluster/cluster.proto
```diff
@@ -0,0 +1,30 @@
...
+
+option java_package = "grakn.protocol.cluster";
+option java_outer_classname = "ClusterProto";
+
+import "protobuf/cluster/discovery.proto";
+
+package grakn.protocol;
+
+service Cluster {
+    rpc cluster_discovery (Discovery.Cluster.Req) returns (Discovery.Cluster.Res);
```
alexjpwalker, line 28: I'm not a huge fan of these method names. In Core, we use an object and an optional verb: `session_open`, `database_delete`, etc. Do you think we should use `cluster_discover` and `database_replicas_discover` (and rename message `Discovery` to `Discover`)?

alexjpwalker (changes requested):

protobuf/cluster/cluster.proto
alexjpwalker: Actually I'm not sure this is the right grouping either. Doesn't it feel like `Cluster` and `DatabaseReplica` should be top-level messages, and `Discover` should be their child?

grpc/java/BUILD
lolski: I'm also not sure yet.

I think it's best to keep the change minimal and keep the current structure, because we're going to see more protocol definitions added for other cluster features such as backup and authentication. That would be the point where we can think about how to re-optimise the structure.

What do you think @alexjpwalker ?

protobuf/cluster/cluster.proto
lolski: > I'm not a huge fan of these method names. In Core, we use an object and an optional verb: session_open, database_delete, etc. Do you think we should use cluster_discover and database_replicas_discover (and rename message Discovery to Discover)?

I agree with this. I think we should update the name to be `cluster_discover` and `database_replicas_discover`.

lolski: > Actually I'm not sure this is the right grouping either. Doesn't it feel like Cluster and DatabaseReplica should be top-level messages, and Discover should be their child?

I kind of like that as well, because then we don't have to introduce too many new top-level terminology. "discovery" is just an operation on existing concepts such as cluster / servers and database.

grpc/java/BUILD
alexjpwalker: Yeah I agree with that.

protobuf/cluster/cluster.proto
```diff
@@ -0,0 +1,32 @@
...
+package grakn.protocol.cluster;
+
+message Cluster {
+    message Discover {
+        message Req {}
+        message Res {
+            repeated string servers = 1;
+        }
+    }
+}
```
alexjpwalker, line 32: hehe
```suggestion
}

```

protobuf/cluster/database.proto
```diff
@@ -0,0 +1,42 @@
...
+
+            message Replica {
+                string address = 1;
+                string database = 2;
+                bool is_leader = 3;
+                int64 term = 4;
+            }
+        }
+    }
+}
```
alexjpwalker, line 42:
```suggestion
}

```

protobuf/cluster/grakn-cluster.proto
```diff
@@ -0,0 +1,31 @@
...
+
+import "protobuf/cluster/cluster.proto";
+import "protobuf/cluster/database.proto";
+
+package grakn.protocol.cluster;
+
+service GraknCluster {
+    rpc cluster_discover (Cluster.Discover.Req) returns (Cluster.Discover.Res);
+    rpc database_discover (Database.Discover.Req) returns (Database.Discover.Res);
+}
```
alexjpwalker, line 31:
```suggestion
}

```

lolski: i thought i was going to the last one faster

alexjpwalker (approved):

PR #110, opened 2021-01-24
Introduce a new transaction type for reading from secondary replicas

lolski:

We've introduced a new Grakn-Cluster-specific transaction type, "read secondary". It is a read-only transaction that may read not only from primary but also from secondary replicas.

- Adds a new option `allow_secondary_replica`

### DIFF
--- a/protobuf/options.proto
+++ b/protobuf/options.proto
@@ -44,7 +44,7 @@ message Options {
     oneof schema_lock_acquire_timeout_opt {
         int32 schema_lock_acquire_timeout_millis = 6;
     }
-    oneof primary_replica_opt {
-        bool primary_replica = 7;
+    oneof allow_secondary_replica_opt {
+        bool allow_secondary_replica = 7;
     }
 }
--- a/protobuf/options.proto
+++ b/protobuf/options.proto
@@ -44,4 +44,7 @@ message Options {
     oneof schema_lock_acquire_timeout_opt {
         int32 schema_lock_acquire_timeout_millis = 6;
     }
+    oneof primary_replica_opt {
+        bool primary_replica = 6;
+    }
 }
--- a/protobuf/options.proto
+++ b/protobuf/options.proto
@@ -45,6 +45,6 @@ message Options {
         int32 schema_lock_acquire_timeout_millis = 6;
     }
     oneof primary_replica_opt {
-        bool primary_replica = 6;
+        bool primary_replica = 7;
     }
 }
--- a/protobuf/transaction.proto
+++ b/protobuf/transaction.proto
@@ -67,6 +67,7 @@ message Transaction {
     enum Type {
         READ = 0;
         WRITE = 1;
+        READ_SECONDARY = 2;
     }
 
     message Open {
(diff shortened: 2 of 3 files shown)


protobuf/transaction.proto
```diff
@@ -67,6 +67,7 @@ message Transaction {
     enum Type {
         READ = 0;
         WRITE = 1;
+        READ_SECONDARY = 2;
```
lolski, line 70: Defining cluster specific type here is not ideal, and I've created an issue for it: https://github.com/graknlabs/protocol/issues/113

alexjpwalker: Why not READ_REPLICA?

lolski: Ah I've answered it here on the other PR: https://github.com/graknlabs/client-java/pull/243/files#r563711191

But let's continue the discussion about the name here so that it's not scattered.

The reasoning to rename it to `READ_SECONDARY` is because it allows you to read from *secondary* replicas, whereas `READ` and `WRITE` reads to the *primary* replica.

The alternative name would be `READ_SECONDARY_REPLICA`?

alexjpwalker: Interesting. I can see the logic.

I'm not familiar enough with this domain to know what would/wouldn't be an appropriate name. `READ_SECONDARY_REPLICA` sounds so damn long. I therefore propose that we go with READ_SECONDARY as you suggest and maybe revisit this after the first release.

alexjpwalker (approved):

alexjpwalker (changes requested):
Updated review to reflect recent discussion

protobuf/transaction.proto
alexjpwalker: After chatting to Haikal we agreed to change `READ_SECONDARY` from being a Transaction type to being an `Option` named `SECONDARY_REPLICA` or similar

lolski: Fixed in https://github.com/graknlabs/protocol/pull/110/commits/68cb20d6e14df0acae1a6aed011562440ce3aa54 and https://github.com/graknlabs/protocol/pull/110/commits/ada1b4c52fcf08b1b7adaea1270192c67067f5ac.

protobuf/options.proto
```diff
@@ -44,4 +44,7 @@ message Options {
     oneof schema_lock_acquire_timeout_opt {
         int32 schema_lock_acquire_timeout_millis = 6;
     }
+    oneof primary_replica_opt {
+        bool primary_replica = 7;
```
alexjpwalker, line 48: I wonder if `primary_replica` is misleading because it could be switched off and the primary replica might still be used. What do you think of the following alternatives:

- `require_primary_replica`
- `allow_secondary_replica`

Or do you think we should stick with `primary_replica` for conciseness?

lolski: Fixed in https://github.com/graknlabs/protocol/pull/110/commits/6d5a5e1b58d91fd001e06b2820363e7194f7fd3f.

alexjpwalker (approved):

PR #112, opened 2021-01-25
Add prefetch option

alexjpwalker:

On executing a Graql query, Grakn automatically streams the first batch of responses back to the client. But, for an `insert` query, they usually don't need those answers. To remedy this, we made `prefetch` a configurable per-query option.

Add prefetch option


flyingsilverfin (approved):

flyingsilverfin:
I think in the scope of this repository, we can probably slim down the PR description. However, its probably fine

PR #118, opened 2021-02-10
Clarify Cluster service methods and introduce "preferred secondary" replica

alexjpwalker:

Cluster database replicas can now be _preferred secondary replicas_. To encapsulate this we added the flag `preferred_secondary` to `cluster.Database.Replica`.

Additionally, we made the Cluster service methods more transparent by naming them `Cluster.Servers` and `Database.Replicas` (previously `Cluster.Discover` and `Database.Discover`)

- Rename `Cluster.Discover` to `Cluster.Servers`
- Rename `Database.Discover` to `Database.Replicas`
- Add `preferred_secondary` to `cluster.Database.Replica`

### DIFF
--- a/protobuf/cluster/database.proto
+++ b/protobuf/cluster/database.proto
@@ -23,20 +23,27 @@ option java_outer_classname = "DatabaseProto";
 package grakn.protocol.cluster;
 
 message Database {
-    message Discover {
+
+    message Replica {
+        string address = 1;
+        string database = 2;
+        Rank rank = 3;
+        int64 term = 4;
+
+        enum Rank {
+            PRIMARY = 0;
+            SECONDARY = 1;
+            BACKUP = 2;
+        }
+    }
+
+    message Replicas {
         message Req {
             string database = 1;
         }
 
         message Res {
             repeated Replica replicas = 1;
-
-            message Replica {
-                string address = 1;
-                string database = 2;
-                bool is_primary = 3;
-                int64 term = 4;
-            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


protobuf/cluster/database.proto
```diff
@@ -23,20 +23,27 @@ option java_outer_classname = "DatabaseProto";
...
+        string address = 1;
+        string database = 2;
+        Rank rank = 3;
+        int64 term = 4;
+
+        enum Rank {
+            PRIMARY = 0;
+            SECONDARY = 1;
+            BACKUP = 2;
+        }
```
alexjpwalker, line 37: The three ranks are:

- `PRIMARY`: This replica should be contacted to initiate primary sessions
- `SECONDARY`: This replica should be contacted to initiate secondary sessions
- `BACKUP`: This replica should only be contacted if a primary or secondary replica fails

protobuf/cluster/database.proto
```diff
@@ -23,20 +23,22 @@ option java_outer_classname = "DatabaseProto";
...
 message Database {
-    message Discover {
+
+    message Replica {
+        string address = 1;
+        string database = 2;
+        bool primary = 3;
+        bool preferred_secondary = 4;
+        int64 term = 5;
+    }
```
alexjpwalker, line 33: I had actually considered introducing an enum field, `rank`. However, this ended up being unclean, because it's possible for the primary replica to **also** be the preferred secondary replica.

lolski (approved):

PR #121, opened 2021-02-13
Set Generic Services option to true

mkschulze:

This pull request sets the generic services Java option to true in order to be able to automatically generate the content of the grakn.proto file with the Protobuf.jl Julia package. 

The [Julia code generator](https://github.com/JuliaIO/ProtoBuf.jl) generates code for generic services if they are switched on for either C++ (cc_generic_services), Python (py_generic_services) or Java (java_generic_services).

Without this option set to true, the file cannot be generated.

We added the required option into the grakn.proto file as described in the instructions of the Protobuf.jl package [here](https://github.com/JuliaIO/ProtoBuf.jl/blob/master/PROTOC.md#generic-services).
It's also described here in the [protobuf language reference](https://developers.google.com/protocol-buffers/docs/reference/java-generated#service).

We added the following line `option java_generic_services = true;` into the grakn.proto file. 
This is the best way because otherwise, we would need to add this option manually or for example with a script, before generating our Julia code files with the Julia code generator. Such a script could break in the future and add further workload for maintenance. It seems to be better to fix this at the source.

Another option may be ([derived from the docs](https://developers.google.com/protocol-buffers/docs/reference/java-generated#service)) to rewrite/rethink the grakn.proto code and tie it to one system, as generic services are deprecated and might infer with plugins at some point.

This solves the issue described here #https://github.com/JuliaIO/ProtoBuf.jl/issues/165


alexjpwalker:
@mkschulze Do we still need this?

We've refactored the codebase - if we do still need this, please add the option to both the new `core_service.proto` and `cluster_service.proto`. I read the docs, seems like something we should have.

mkschulze:
@alexjpwalker yes, it's still required. I added the option to both files now. thank you.

haikalpribadi (approved):

haikalpribadi:
Thanks, @mkschulze ! 👍

PR #126, opened 2021-03-04
Add 'type' field to Thing

alexjpwalker:

The `Thing` message now contains a new field, `type`, which holds the concept type. As a result of this change, `value_type` is now redundant (equivalent to `type.value_type`) and `encoding` can be inferred from `type.encoding`, so both fields have been deleted.

Add 'type' field to Thing message; delete `value_type` and `encoding`

### DIFF
--- a/protobuf/concept.proto
+++ b/protobuf/concept.proto
@@ -105,9 +105,8 @@ message Concept {
 message Thing {
 
     bytes iid = 1;
-    Encoding encoding = 2;
-    AttributeType.ValueType value_type = 3;
-    Attribute.Value value = 4;
+    Type type = 2;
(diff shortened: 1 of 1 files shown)


protobuf/concept.proto
```diff
@@ -105,9 +105,8 @@ message Concept {
 message Thing {
 
     bytes iid = 1;
-    Encoding encoding = 2;
-    AttributeType.ValueType value_type = 3;
-    Attribute.Value value = 4;
```
alexjpwalker, line 110: `value_type` is now redundant (equivalent to `type.value_type`) and `encoding` can be inferred from `type.encoding`.

flyingsilverfin (approved):

PR #129, opened 2021-03-15
Architecture Refactor

haikalpribadi:

The previous definition of Grakn RPC protocol did not differentiate the fact that, for the Transaction Bidirectional Stream, some server responses are complete responses for their associated queries, and others are a partial list of a response which are streamed consecutively to make up a full response incrementally. This makes the protocol not explicitly clear to the developer implementing Grakn clients, without having the internal knowledge of how Grakn server responses are designed. We have now refined the Transaction Bidirectional Stream protocol, to express the server behaviour more explicitly. The Transaction Bidirectional Stream now sends "client" messages and receives back "server" messages. "Server" messages are defined to either be a complete response or a partial response. Additionally, we have refined the variable names for some other message definitions.

This PR is followed by https://github.com/graknlabs/client-java/pull/302 in client-java, and another one in Grakn Core (to be created).

### DIFF
--- a/protobuf/cluster/cluster.proto
+++ b/protobuf/cluster/cluster.proto
@@ -26,7 +26,7 @@ message Cluster {
     message Servers {
         message Req {}
         message Res {
-            repeated string servers = 1;
+            repeated string addresses = 1;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


alexjpwalker (approved):
Looks good to me

lolski (approved):

PR #130, opened 2021-03-19
Refactor protocol file structure and message names

haikalpribadi:

The previous file structure did not communicate clearly the architecture of our protocol definitions. We have Grakn Core specific protocol definitions, as well as Grakn Cluster specific ones. These protocol files and their message names were also not named accurately based on their domains. For example, `database.proto` existed twice in Cluster and Core protobuf packages. We also split messages that should be two different message types, such as `ClusterDatabase` and `ClusterDatabaseManager`.

- Renamed `//protobuf/cluster/cluster.proto` to `cluster_server.proto`
- Renamed `//cluster/database.proto` to `//cluster/cluster_database.proto`
- Renamed `grakn.protocol.cluster.Database` to `ClusterDatabase`
- Split `ClusterDatabaseManager` out from `ClusterDatabase`
- Moved `//protobuf/cluster` out to its own package `//cluster`
- Split `//protobuf` into `//common` and `//core
- Set package address of all protocol definitions and their generated files to `grakn.protocol`


grpc/nodejs/BUILD
```diff
@@ -41,30 +41,30 @@ nodejs_binary(
...
 
 ts_grpc_compile(
     name = "protocol-cluster",
     deps = [
-        "//protobuf/cluster:cluster-proto",
-        "//protobuf/cluster:database-proto",
-        "//protobuf/cluster:grakn-cluster-proto",
+        "//cluster:server-proto",
+        "//cluster:database-proto",
+        "//cluster:service-proto", # TODO: do we need this?
```
lolski, line 64: Let's remove it since it's already declared in `grpc_deps`

cluster/cluster_database.proto
```diff
@@ -17,36 +17,39 @@
...
+    string name = 1;
+    repeated Replica replicas = 2;
+
+    message Replica {
+        string address = 1;
+        bool primary = 2;
+        bool preferred = 3;
+        int64 term = 4;
+    }
+}
```
lolski, line 55: There's a missing newline but I think it's fine right? Newline is not required, I've never seen it cause a build to be broken

lolski (approved):

grpc/python/BUILD
```diff
@@ -57,7 +57,7 @@ python_repackage(
 python_repackage(
     name = "protocol-cluster-pkg",
     src = ":protocol-cluster-src",
-    src_package = "protobuf.cluster",
+    src_package = "cluster",
```
alexjpwalker, line 60: I'm not 100% sure if this will work with the current implementation of `python_repackage` (not sure if it replaces `import cluster` or just `cluster`) so we should test that ASAP

alexjpwalker (approved):

grpc/nodejs/BUILD
alexjpwalker: I don't know if that will work - if we do remove it, we should re-test it.

cluster/cluster_database.proto
haikalpribadi: We should always end a file with a newline regardless of build not complaining, @lolski , so I'll update this

grpc/nodejs/BUILD
haikalpribadi: I'm with @alexjpwalker here @lolski . We should remove it when we can test it. So that will have to be done manually from client-nodejs when any one of us gets there.

PR #131, opened 2021-03-22
Implement Explanations in RPC

flyingsilverfin:
We add Explanations, and the explain() query endpoint to the protocol. This enables all clients to retrieve all available explanations for a particular explainable concept map.

* Implement `Explainable` and `Explainables` as part of `Answer` protocol
* Implement `Explanation` as part of `Logic` protocol
* Add `Explain.Req` and `Explain.Res` as part of the explain QueryManager endpoing

### DIFF
--- a/common/answer.proto
+++ b/common/answer.proto
@@ -26,7 +26,24 @@ package grakn.protocol;
 
 message ConceptMap {
     map<string, Concept> map = 1;
-    string pattern = 2;
+    Explainables explainables = 2;
+}
+
+message Explainables {
+    map<string, Explainable> explainable_relations = 1;
+    map<string, Explainable> explainable_attributes = 2;
+    repeated ExplainableOwnership explainable_ownerships = 3;
+}
+
+message ExplainableOwnership {
+    string owner = 1;
+    string attribute = 2;
+    Explainable explainable = 3;
+}
+
+message Explainable {
+    string conjunction = 1;
+    int64 id = 2;
 }
 
 message ConceptMapGroup {
--- a/common/answer.proto
+++ b/common/answer.proto
@@ -30,15 +30,13 @@ message ConceptMap {
 }
 
 message Explainables {
-    map<string, Explainable> explainable_relations = 1;
-    map<string, Explainable> explainable_attributes = 2;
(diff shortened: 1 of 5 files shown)


common/logic.proto
```diff
@@ -106,3 +109,14 @@ message Rule {
         message Res {}
     }
 }
+
+message Explanation {
+    Rule rule = 1;
+    map<string, VarsList> varMapping = 2; // TODO perhaps this should be a "context" and contain more information about source
```
alexjpwalker, line 115: We need to use `snake_case` consistently everywhere for protocol field names - Java handles mixed casing relatively gracefully, but Python and JS don't.
```suggestion
    map<string, VarsList> var_mapping = 2; // TODO perhaps this should be a "context" and contain more information about source
```
(Other examples elsewhere in the PR)

alexjpwalker (approved):

common/answer.proto
```diff
@@ -26,7 +26,24 @@ package grakn.protocol;
...
+
+message ExplainableOwnership {
+    string owner = 1;
+    string attribute = 2;
+    Explainable explainable = 3;
+}
+
+message Explainable {
+    string conjunction = 1;
+    int64 id = 2;
```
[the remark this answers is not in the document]
haikalpribadi: let's try `map<string, map<string, Explainable>>` if we can do it in GRPC

common/logic.proto
```diff
@@ -106,3 +109,14 @@ message Rule {
         message Res {}
     }
 }
+
+message Explanation {
+    Rule rule = 1;
+    map<string, VarsList> var_mapping = 2; // TODO perhaps this should be a "context" and contain more information about source
+    ConceptMap then_answer = 3;
+    ConceptMap when_answer = 4;
```
haikalpribadi, line 117: `conclusion` and `condition` are better names

common/logic.proto
```diff
@@ -106,3 +109,14 @@ message Rule {
...
+
+    message VarsList {
```
haikalpribadi, line 119: This should be `VarList`

haikalpribadi (changes requested):

haikalpribadi (approved):

PR #135, opened 2021-04-07
Store request IDs as bytes; add 'inferred' to Thing

### LINKED ISSUE #134: Store UUIDs as bytes, not strings
## Problem to Solve

UUID representation as string is inefficient.

## Proposed Solution

Store Session IDs and Request IDs (and any other UUIDs) as bytes, not strings.

alexjpwalker:

Previously the field `req_id` of a transaction request was stored as a string, which is an inefficient representation of a UUID. We've changed it to bytes.

Also, we deleted `IsInferred` and added the `inferred` flag to `Thing`, allowing clients to check if a Thing is inferred without an additional network roundtrip.

Store request IDs as bytes, not strings (fixes #134 )
Add 'inferred' to Thing (fixes https://github.com/graknlabs/client-java/issues/314 )


.bazelrc
```diff
@@ -15,9 +15,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-build --incompatible_strict_action_env --javacopt='--release 8'
+build --incompatible_strict_action_env --java_language_version=11 --javacopt='--release 11'
```
alexjpwalker, line 18: See https://github.com/graknlabs/grakn/pull/6281 - after installing the newest Bazel (by pulling `grakn` and building it locally) I was no longer able to build `protocol`. I fixed it by upgrading Bazel and Java in protocol.

common/concept.proto
```diff
@@ -107,19 +107,19 @@ message Thing {
...
-            Thing.GetHas.Req thing_get_has_req = 103;
-            Thing.SetHas.Req thing_set_has_req = 104;
-            Thing.UnsetHas.Req thing_unset_has_req = 105;
-            Thing.GetRelations.Req thing_get_relations_req = 106;
-            Thing.GetPlaying.Req thing_get_playing_req = 107;
+            Thing.GetHas.Req thing_get_has_req = 102;
+            Thing.SetHas.Req thing_set_has_req = 103;
+            Thing.UnsetHas.Req thing_unset_has_req = 104;
+            Thing.GetRelations.Req thing_get_relations_req = 105;
+            Thing.GetPlaying.Req thing_get_playing_req = 106;
```
alexjpwalker, line 122: Note that moving these around breaks compatibility with older clients - maybe this is a practice we want to abolish going forward?

haikalpribadi (approved):

PR #136, opened 2021-04-23
Rename Grakn to TypeDB

alexjpwalker:

We renamed Grakn to TypeDB. All occurrences of "Grakn" have been replaced with "TypeDB".

Rename Grakn to TypeDB


haikalpribadi:
There were lots of mistakes in this one, such as `TypeDB Labs`, @alexjpwalker , and we want to change the java package address to globally start with `com.vaticle.` now. I've proceeded with with the refactor without this PR.

PR #141, opened 2021-05-26
Added user management protocol for TypeDB Cluster

lolski:

We have added the user management protocol for TypeDB Cluster. It will allow TypeDB clients to perform user management operations such as adding and deleting a user.

- Add `users_create`, `users_contains`, `users_all`, and `user_delete` endpoints in `TypeDBCluster` service.


lolski:
The authentication feature requires the user credential to be sent in requests. The credential will be included as gRPC metadata, and therefore there's no need to add it as a field in the protobuf definition.

PR #144, opened 2021-09-02
Token renewal protocol for TypeDB Cluster

lolski:

We have introduced a new protocol for performing authentication token renewal in TypeDB Cluster.

1. Added a new service `user_token_renew`
2. Added `Renew` request and response object

### DIFF
--- a/cluster/cluster_service.proto
+++ b/cluster/cluster_service.proto
@@ -40,6 +40,7 @@ service TypeDBCluster {
     // User API
     rpc user_password (ClusterUser.Password.Req) returns (ClusterUser.Password.Res);
     rpc user_delete (ClusterUser.Delete.Req) returns (ClusterUser.Delete.Res);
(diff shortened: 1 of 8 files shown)


cluster/cluster_service.proto
```diff
@@ -41,6 +43,9 @@ service TypeDBCluster {
     rpc user_password (ClusterUser.Password.Req) returns (ClusterUser.Password.Res);
     rpc user_delete (ClusterUser.Delete.Req) returns (ClusterUser.Delete.Res);
 
+    // User Token API
+    rpc user_token_renew (ClusterUserToken.Renew.Req) returns (ClusterUserToken.Renew.Res);
```
alexjpwalker, line 47: Should this perhaps be `SessionToken`?

lolski: A "user token" isn't only used to authenticate session. For instance, it is also used to authenticate database management operations.

Given that one user will have at most one token per server instance, I think the name user token is apt.

alexjpwalker (approved):

PR #150, opened 2022-01-11
Use token-based authentication for npm deployment

alexjpwalker:

The npm deployment infrastructure now uses token-based authentication instead of username/password authentication.

See https://github.com/vaticle/bazel-distribution/pull/330 for an explanation of why this PR is necessary.

- Update dependencies (to make `deploy_npm` use token authentication) and automation.yml (to pass the token in)

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -64,7 +64,7 @@ build:
         sudo apt update -y
         sudo apt install -y expect
         export DEPLOY_NPM_EMAIL=$REPO_VATICLE_EMAIL
-        export DEPLOY_NPM_TOKEN=$REPO_VATICLE_TOKEN
+        export DEPLOY_NPM_TOKEN=$REPO_VATICLE_NPM_TOKEN
         bazel run --define version=$(git rev-parse HEAD) //grpc/nodejs:deploy-npm -- snapshot
       dependencies: [build, build-dependency]
     deploy-pip-snapshot:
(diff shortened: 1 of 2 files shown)


.grabl/automation.yml
```diff
@@ -63,9 +63,7 @@ build:
         wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
```
alexjpwalker, line 63: Update CI scripts: we no longer need username/email/password for authentication when deploying the npm package - we just need a reusable auth token.

The token used depends on which repo we're deploying to, snapshot or release.

dependencies/vaticle/repositories.bzl
```diff
@@ -20,6 +20,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
```
alexjpwalker, line 20: Update to dependencies that gives us access to the token-based `deploy_npm` rule

dependencies/vaticle/repositories.bzl
```diff
@@ -21,5 +21,5 @@ def vaticle_dependencies():
     git_repository(
```
alexjpwalker, line 21: Update to dependencies that gives us access to the token-based `deploy_npm` rule

lolski (approved):

PR #152, opened 2022-01-17
Update deploy_npm build tool

alexjpwalker:

We've updated the `deploy_npm` build tool, as the previous version was not fully reliable in how it passed the auth token to the tool.

Update dependencies


dependencies/vaticle/repositories.bzl
```diff
@@ -20,6 +20,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
```
alexjpwalker, line 20: We've updated the deploy_npm build tool, as the previous version was not fully reliable in how it passed the auth token to the tool.

lolski (approved):

haikalpribadi (approved):

PR #153, opened 2022-01-18
Add missing CI dependency in deploy-pip-snapshot

alexjpwalker:

`deploy-pip-snapshot` was incorrectly being run concurrently with `build` in CI. So we've now marked `build` as a dependency of `deploy-pip-snapshot`.

Add missing CI dependency in deploy-pip-snapshot


.grabl/automation.yml
```diff
@@ -80,6 +80,7 @@ build:
         export DEPLOY_PIP_USERNAME=$REPO_VATICLE_USERNAME
```
alexjpwalker, line 80: `deploy-pip-snapshot` was incorrectly being run concurrently with `build` in CI. So we've now marked build as a dependency of `deploy-pip-snapshot`.

haikalpribadi (approved):

PR #168, opened 2022-11-15
Add Rust distribution targets and their CI workflows

alexjpwalker:

We added crate distribution targets (`assemble_crate` and `deploy_crate`) for Rust, and added CI workflows for deploying snapshot and release crates. 

We recently tried adding an `assemble_crate` rule to our existing Rust proto codebase to distribute Rust proto bindings, but it didn't work - it didn't contain any protobuf source code, because the `assemble_crate` rule doesn't support Cargo build scripts.

But instead of actually adding that support, we had a rethink of the build architecture. The old idea was (following Tonic's own docs) for `typedb-protocol` to come bundled with the `.proto` sources and a Cargo build script that would allow the end-user to compile the protobufs on-the-fly. This would require them to have `protoc` installed in their environment, and would dramatically increase their compilation time and overhead.

A better option would be to bundle the generated Rust sources. So we've created the `rust_tonic_compile` rule that enables just that in https://github.com/vaticle/dependencies/pull/391. Now, both `rust_library` and `assemble_crate` rules can depend on `typedb_protocol` seamlessly, so we can depend on `typedb_protocol` either through Bazel or Cargo.

As this PR stabilises the release process, we've gone ahead and added the relevant CI workflows: `deploy-crate-snapshot` and `deploy-crate-release`.


dependencies/vaticle/repositories.bzl
```diff
@@ -20,6 +20,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "5be6d949ca1e04e4179ea6acb3432be713b9dfb8", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/alexjpwalker/dependencies",
+        commit = "3244740dd87cb91306fe0dc3418f57ccc640cd0e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
```
alexjpwalker, line 24: Update to `vaticle` once https://github.com/vaticle/dependencies/pull/383 is merged

grpc/rust/BUILD
```diff
@@ -16,13 +16,15 @@
 #
```
alexjpwalker, line 16: The previous build architecture, with a `cargo_build_script` at its core, would not have led to a good UX. The idea was (following Tonic's own docs) for `typedb-protocol` to come bundled with the `.proto` sources and a build script that would allow the end-user to compile the protobufs on-the-fly. This would require them to have `protoc` installed in their environment, and would dramatically increase their compilation time and overhead.

Instead, we needed a way to bundle the generated Rust sources. First, we created a `rust_binary` that ran the build script. But this was a `bazel run` target; we needed a `build` target. So now, we've created a rule, `rust_tonic_compile`, that internally runs this `rust_binary` (similarly to how many of our rules run Kotlin binaries), and exposes the outputs. Any `rust_library` can then depend on these generated sources. This is similar to the behaviour of the `antlr` rule, and the API is similar to Java and Python (`java_grpc_compile` and `python_grpc_compile` respectively)

grpc/rust/build.rs
```diff
@@ -1,43 +0,0 @@
-//
```
alexjpwalker, line 1: We actually refactored (moved) this file to the `grpc/rust/compile` package (but Git doesn't know because we also massively changed the code)

grpc/rust/compile/BUILD
```diff
@@ -0,0 +1,33 @@
+#
```
alexjpwalker, line 1: This package contains the `rust_tonic_compile` rule and the `rust_binary` that powers it.

Ideally, we'd get this rule from an external repo such as `rules_rust` or `rules_proto`, but they don't have `tonic` implementations yet.

We may want to relocate this rule to another repo so it can be used by other Rust proto repos.

grpc/rust/compile/compile.rs
```diff
@@ -0,0 +1,28 @@
+//
```
alexjpwalker, line 1: We now pass in all the `tonic_build` configuration via environment variables, which are set by the `rust_tonic_compile` rule.

grpc/rust/compile/tonic_compile.bzl
```diff
@@ -0,0 +1,70 @@
+#
```
alexjpwalker, line 1: The rule implementation is quite straightforward:

1. Define the inputs for the compile script, namely all `.proto` sources and the `protoc` executable.
2. Define the outputs of the compile script. One output is generated for each unique `package` defined in our `.proto` files. Bazel can't (easily!) get these package names, so the user is required to supply them as a parameter.
3. Run the compile script and expose the outputs as source files that any `rust_library` can depend on.

(The `rust_library` is then expected to depend on the `tonic` and `prost` crates, otherwise it will fail to compile. This is in line with Java, where we expect our `java_library` to depend on the correct gRPC + Protobuf JARs.)

grpc/rust/lib.rs
```diff
@@ -15,4 +15,4 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 //
 
-tonic::include_proto!("typedb.protocol");
+include!("typedb.protocol.rs");
```
alexjpwalker, line 18: The old form `tonic::include_proto!` is basically designed for Cargo build scripts; it depends on an environment variable (`OUT_DIR`) that they set. The `include!` macro allows us to easily reference generated code provided we know its exact location.

grpc/rust/BUILD
```diff
@@ -36,30 +38,42 @@ cargo_build_script(
...
+        "lib.rs",
+    ],
+    deps = [
         "@vaticle_dependencies//library/crates:tonic",
         "@vaticle_dependencies//library/crates:prost",
     ],
     tags = [
         "crate-name=typedb-protocol",
     ],
+    version = "0.1.2",
```
alexjpwalker, line 57: We released version `0.1.2` manually, so there's no need to re-release this. However, we should add the CI release scripts.

grpc/rust/compile/tonic_compile.bzl
```diff
@@ -0,0 +1,70 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+def _rust_tonic_compile_impl(ctx):
+    inputs = ctx.attr.protoc.files.to_list()
+    proto_paths_str = ""
+    for src in ctx.attr.srcs:
+        inputs.append(src[ProtoInfo].direct_sources[0])
+        proto_paths_str = proto_paths_str + src[ProtoInfo].direct_sources[0].path + ";"
```
dmitrii-ubskii, line 23: This could be simplified to
```suggestion
    sources = [src[ProtoInfo].direct_sources[0] for src in ctx.attr.srcs]
    inputs = ctx.attr.protoc.files.to_list() + sources
    proto_paths_str = ";".join([src.path for src in sources])
```
The last line could be inlined in the `ctx.actions.run()` call as well.

grpc/rust/compile/tonic_compile.bzl
```diff
@@ -0,0 +1,70 @@
...
+
+    outputs = []
+    for package in ctx.attr.packages:
+        outputs.append(ctx.actions.declare_file("{}.rs".format(package)))
```
dmitrii-ubskii, line 27: Also consider list comprehension here:
```suggestion
    outputs = [ctx.actions.declare_file("{}.rs".format(package)) for package in ctx.attr.packages]
```

dmitrii-ubskii (changes requested):
Couple pythonic Starlark style suggestions.

grpc/rust/compile/tonic_compile.bzl
alexjpwalker: Yeah, good point. To be honest I couldn't remember if Starlark supported list comprehensions so I lazily just went with loops. Updated to use comprehensions.

dmitrii-ubskii (approved):

.factory/automation.yml
```diff
@@ -79,6 +79,15 @@ build:
         export DEPLOY_PIP_PASSWORD=$REPO_VATICLE_PASSWORD
         bazel run --define version=$(git rev-parse HEAD) //grpc/python:deploy-pip -- snapshot
       dependencies: [build, build-dependency]
+    deploy-crate-snapshot:
+      filter:
+        owner: vaticle
+        branch: master
+      image: vaticle-ubuntu-22.04
+      command: |
+        export DEPLOY_CRATE_TOKEN=$REPO_VATICLE_CRATES_TOKEN
```
alexjpwalker, line 88: `CRATE` = the package type (similar to `pip`, `npm`)
`CRATES` = the registry type (similar to `PyPi`, `npm`)

dependencies/vaticle/repositories.bzl
```diff
@@ -21,5 +21,5 @@ def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
         remote = "https://github.com/vaticle/dependencies",
-        commit = "5be6d949ca1e04e4179ea6acb3432be713b9dfb8", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        commit = "3a355418fad6b632ab6cd809945296b6756ea076", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
```
alexjpwalker, line 24: New version of `dependencies` contains fixes for building on M1 Macs.

grpc/rust/compile/compile.bzl
```diff
@@ -0,0 +1,62 @@
+#
```
alexjpwalker, line 1: Algorithm:

1. Collect the protos (`.proto` sources) and the `protoc` executable, declaring them as `inputs`;
2. Declare one `output` file for each unique Protobuf `package` in the proto sources (for Protocol, there is just one package, `typedb.protocol`)
3. Run a Rust binary that uses the `tonic_build` crate to build the Rust gRPC + Protobuf bindings.

dependencies/vaticle/repositories.bzl
```diff
@@ -20,6 +20,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "5be6d949ca1e04e4179ea6acb3432be713b9dfb8", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/alexjpwalker/dependencies",
+        commit = "1bf0ec950a59ad170c826f9fe163701c2e0b2233", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
```
alexjpwalker, line 24: Update to vaticle before merging.

grpc/rust/BUILD
```diff
@@ -16,13 +16,15 @@
...
+load("@vaticle_dependencies//distribution:deployment.bzl", "deployment")
 load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("@vaticle_dependencies//builder/grpc/rust:compile.bzl", "rust_tonic_compile")
 
-cargo_build_script(
+rust_tonic_compile(
     name = "typedb_protocol_src",
-    srcs = ["build.rs"],
-    data = [
+    packages = ["typedb.protocol"],
```
alexjpwalker, line 26: The list of Protobuf packages in the `.proto` files. `rust_tonic_compile` needs this info to know what output files it is expected to generate.

lolski (approved):

haikalpribadi (approved):

PR #177, opened 2023-01-16
Migrate to crate-universe

dmitrii-ubskii:

We update dependencies and the Bazel build files to reflect the new style of crate imports using crate-universe instead of cargo-raze.


grpc/rust/BUILD
```diff
@@ -45,11 +45,10 @@ rust_library(
     name = "typedb_protocol",
     srcs = [
         ":typedb_protocol_src",
-        "lib.rs",
```
dmitrii-ubskii, line 48: This file is not necessary for the library compilation, and mixing generated and non-generated sources breaks rust library compilation as an external dependency.

alexjpwalker (approved):

PR #179, opened 2023-02-27
Password policy and user management improvement

jamesreprise:

We've added two protobuf messages and modified one, enabling us to provide password policy features to TypeDB Cluster. See the PR for further information: https://github.com/vaticle/typedb-cluster/pull/456

We've added:
* `users_password_set`, a message carrying the information for an administrator to overwrite any other user's existing password.
* `users_get`, that returns the user and associated information (currently only the username and the number of days before said user's password expires.)
* The `User` message, which includes all the information returned by `users_get` and `users_all`.

We've modified:
* `users_delete`, moving the method to the user manager.
* `user_password_update`, splitting it into `users_password_set` for administrators and this message that allows user's to update their own passwords.

### DIFF
--- a/cluster/cluster_service.proto
+++ b/cluster/cluster_service.proto
@@ -40,6 +40,8 @@ service TypeDBCluster {
 
     // User API
     rpc user_password (ClusterUser.Password.Req) returns (ClusterUser.Password.Res);
+    rpc user_password_admin (ClusterUser.PasswordAdmin.Req) returns (ClusterUser.PasswordAdmin.Res);
+    rpc user_expiry_days (ClusterUser.ExpiryDays.Req) returns (ClusterUser.ExpiryDays.Res);
     rpc user_token (ClusterUser.Token.Req) returns (ClusterUser.Token.Res);
     rpc user_delete (ClusterUser.Delete.Req) returns (ClusterUser.Delete.Res);
 
--- a/cluster/cluster_service.proto
+++ b/cluster/cluster_service.proto
@@ -36,14 +36,14 @@ service TypeDBCluster {
     // User Manager API
     rpc users_contains (ClusterUserManager.Contains.Req) returns (ClusterUserManager.Contains.Res);
     rpc users_create (ClusterUserManager.Create.Req) returns (ClusterUserManager.Create.Res);
+    rpc users_delete (ClusterUserManager.Delete.Req) returns (ClusterUserManager.Delete.Res);
     rpc users_all (ClusterUserManager.All.Req) returns (ClusterUserManager.All.Res);
+    rpc users_password_set (ClusterUserManager.PasswordSet.Req) returns (ClusterUserManager.PasswordSet.Res);
+    rpc users_get (ClusterUserManager.Get.Req) returns (ClusterUserManager.Get.Res);
 
     // User API
-    rpc user_password (ClusterUser.Password.Req) returns (ClusterUser.Password.Res);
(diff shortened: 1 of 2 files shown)


cluster/cluster_service.proto
```diff
@@ -40,6 +40,8 @@ service TypeDBCluster {
 
     // User API
     rpc user_password (ClusterUser.Password.Req) returns (ClusterUser.Password.Res);
+    rpc user_password_admin (ClusterUser.PasswordAdmin.Req) returns (ClusterUser.PasswordAdmin.Res);
+    rpc user_expiry_days (ClusterUser.ExpiryDays.Req) returns (ClusterUser.ExpiryDays.Res);
```
flyingsilverfin, line 44: let's use:
`user_password_set` - the admin can set another user's password
`user_password_update` - a user can update new users


The final set of APIs will be:
```
    // User Manager API -- these are mostly available to only admins
    rpc users_contains (ClusterUserManager.Contains.Req) returns (ClusterUserManager.Contains.Res);
    rpc users_create (ClusterUserManager.Create.Req) returns (ClusterUserManager.Create.Res);
** rpc users_delete (ClusterUserManager.Delete.Req) returns (ClusterUserManager.Delete.Res);
    rpc users_all (ClusterUserManager.All.Req) returns (ClusterUserManager.All.Res); 
** rpc users_password_set (ClusterUserManager.PasswordSet.Req) returns (ClusterUserManager.PasswordSet.Res);
** rpc users_get .... this one will be available to each user OR admins


    // User API -- these are accessible to the logged in user
** rpc user_password_update(ClusterUser.PasswordUpdate.Req) returns (ClusterUser.PasswordUpdate.Res);
    rpc user_token (ClusterUser.Token.Req) returns (ClusterUser.Token.Res);
```

flyingsilverfin: The `User` object that is returned from `_get` and `_all` contains the fields:
```
name: string
passwordExpiryDays: int
```

cluster/cluster_user.proto
```diff
@@ -68,12 +93,12 @@ message ClusterUser {
...
+}
 
-    message Delete {
-        message Req {
-            string username = 1;
-        }
-        message Res {}
+message User {
+    string username = 1;
+    oneof password_expiry_opt {
```
flyingsilverfin, line 100: if you check our other protocol definitions I believe you won't find that we label optional fields with `_opt` exception the `Options` ;) 

I would just end up with:
```
message User {
    string username = 1;
    oneof password_expiry {
        int64 password_expiry_days = 2;
    }
}
```

Note that you should be consistent with your casing! We should use `_` case everywhere

cluster/cluster_user.proto
```diff
@@ -41,20 +41,45 @@ message ClusterUserManager {
...
 
 message ClusterUser {
 
-    message Password {
+    message PasswordUpdate {
         message Req {
             string username = 1;
-            string password = 2;
+            string oldPassword = 2;
+            string newPassword = 3;
```
flyingsilverfin, line 82: `password_old` and `password_new`

jamesreprise: Implemented and propagated.

cluster/cluster_user.proto
jamesreprise: Agreed, implemented.

flyingsilverfin (approved):
Good! Let's merge this when we merge all the other ones all together, so just hold off for a bit :)

PR #182, opened 2023-04-11
Introduce message types to encode annotations

flyingsilverfin:

We replace usages of `boolean` for key arguments with a message type called `Annotation` - this gives us a natural place to extend the set/capabilities of the annotations in the future. The new message type is used to implement the new "unique" annotation.

Introduce `Type.Annotation`, which allows carrying either `Key` or `Unique` annotation. This message type is used through the Concept API protocol specification and replaces all usages of boolean `key` arguments.

### DIFF
--- a/common/concept.proto
+++ b/common/concept.proto
@@ -594,7 +594,7 @@ message ThingType {
             oneof filter {
                 AttributeType.ValueType value_type = 1;
             }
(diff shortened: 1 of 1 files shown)


flyingsilverfin:
Verbal approval from Haikal

PR #183, opened 2023-04-18
Api rearchitecture

dmitrii-ubskii:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

{ In the form of a paragraph (only use bullet points if strictly necessary), please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. This section will be automatically compiled into the release notes, so please:
  - describe the impact of the change in this PR to the _user_ of this repository (e.g. end user, contributor, developer).
  - describe the new product behaviour in _present tense_, and the old behaviour and how it's been changed in _past tense_.
  - Use the _Royal We_: _"We"_ made changes, not _"I"_ made changes. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above. Please describe every method, class and package, by explaining:
  - its responsibility, 
  - how it's expected to behave, and 
  - how it relates to the adjacent methods/classes/packages it interacts with. 

This would allow the reviewer to understand your intentions in the code much better. If you're adding new classes, make sure these explanations are also included in the class header comments. Last but not least, please reference the GitHub issues to be automatically closed, such like 'closes #number'. }


proto/concept.proto
```diff
@@ -0,0 +1,979 @@
...
+        oneof req {
+            GetEntityType.Req get_entity_type_req = 1;
+            GetRelationType.Req get_relation_type_req = 2;
+            GetAttributeType.Req get_attribute_type_req = 3;
+            GetEntity.Req get_entity_req = 4;
+            GetRelation.Req get_relation_req = 5;
+            GetAttribute.Req get_attribute_req = 6;
+            PutEntityType.Req put_entity_type_req = 7;
+            PutRelationType.Req put_relation_type_req = 8;
+            PutAttributeType.Req put_attribute_type_req = 9;
```
dmitrii-ubskii, line 37: Group Type requests

dependencies/vaticle/repositories.bzl
```diff
@@ -20,6 +20,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "66596a14ec39921e04aa336af8439962d4ee0005", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/dmitrii-ubskii/dependencies",
+        branch = "api-rearchitecture", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
```
dmitrii-ubskii, line 24: Pending https://github.com/vaticle/dependencies/pull/442

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+            Relation.GetRelating.Req relation_get_relating_req = 204;
+
+            Attribute.GetOwners.Req attribute_get_owners_req = 300;
+        }
+    }
+
+    message Res {
+        oneof res {
+            Thing.Delete.Res thing_delete_res = 100;
+            Thing.SetHas.Res thing_set_has_res = 102;
```
flyingsilverfin: Let's make the numbering blocks sequential, I think the naming lining up is enough.
eg. this section should be
```
... = 100
... = 101
... = 102

... = 200
... = 201
...
```

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+    message Delete {
+        message Req {}
+        message Res {}
+    }
+
+    message GetHas {
+        message Req {
+            oneof filter {
+                AttributeTypes attribute_types = 1;
+                Annotations annotation_filter = 2;
```
flyingsilverfin, line 229: Let's make an issue to remove the `oneof` here, it's not necessary IMO. We'll have to update server & clients with this change. Mark it as `good first issue` and assign it to Dmitrii M for future work.

flyingsilverfin: Note in the issue that the `repeated` can then be brought back inline.

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+        message Req {
+            repeated RoleType role_types = 1;
+        }
+        message ResPart {
+            repeated Thing things = 1;
+        }
+    }
+
+    message GetPlayersByRoleType {
+        message RoleTypeWithPlayer {
```
flyingsilverfin: Let's turn this into `Relation.RolePlayer` and replace the usages of their pair in all of the Relation sub-messages that match exactly (`AddPlayer.Request, RemovePlayer.Reques`- these become renamed `AddRolePlayer` and `RemoveRolePlayer`) etc.

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+            sint64 long = 3;
+            double double = 4;
+            // time since epoch in milliseconds
+            sint64 date_time = 5;
+        }
+    }
+
+    message GetOwners {
+        message Req {
+            optional ThingType filter = 1;
```
flyingsilverfin, line 348: since this is called `filter`, maybe you want to use `_filter` for the suffixes in `ThingType.getHas`?

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+            repeated Thing things = 1;
+        }
+    }
+}
+
+message Type {
+    message Req {
+        oneof req {
+            RoleType.Req role_type_req = 1;
+            ThingType.Req thing_type_req = 2;
```
flyingsilverfin, line 360: Let's make `thing_type` requests come first since they're more 'primary' in our minds than role types (apply throughout)

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+        }
+    }
+}
+
+message ThingType {
+    oneof type {
+        EntityType entity_type = 1;
+        RelationType relation_type = 2;
+        AttributeType attribute_type = 3;
+        ThingType.Root root_thing_type = 10;
```
flyingsilverfin, line 510: `thing_type_root`

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+    message UnsetAbstract {
+        message Req {}
+        message Res {}
+    }
+
+    message GetOwns {
+        message Req {
+            optional AttributeType.ValueType filter = 1;
+            Type.Transitivity transitivity = 2;
+            repeated Type.Annotation annotations = 3;
```
flyingsilverfin, line 646: I think these should all be suffixed `_filter`

flyingsilverfin: Interesting that this supports _all_ of them, but `get_has` only supports one at a time huh

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+    message UnsetOwns {
+        message Req {
+            AttributeType attribute_type = 1;
+        }
+        message Res {}
+    }
+
+    message GetPlays {
+        message Req {
+            optional Type.Transitivity transitivity = 1;
```
flyingsilverfin, line 680: Why is transitivity optional here but not elsewhere? Lets be consistent at each place

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+    message GetRelatesOverridden {
+        message Req {
+            string label = 1;
+        }
+        message Res {
+            optional RoleType role_type = 1;
+        }
+    }
+
+    message SetRelates {
```
flyingsilverfin, line 849: I wonder if this should be `createRelates/deleteRelates` since we don't create RoleType any other way, is that right?

Bit tricky since it should be `createRole` and `deleteRole` or something like that in reality. 

We can leave it for now since it's no different to before. Worth a thought though.

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+        message Req {}
+        message ResPart {
+            repeated AttributeType attribute_types = 1;
+        }
+    }
+
+    message GetSubtypes {
+        message Req {
+            Type.Transitivity transitivity = 1;
+            optional ValueType value_type = 2;
```
flyingsilverfin, line 922: I think these should also `be _filter` right?

proto/concept.proto
```diff
@@ -0,0 +1,963 @@
...
+        }
+        message ResPart {
+            repeated AttributeType attribute_types = 1;
+        }
+    }
+
+    message GetInstances {
+        message Req {
+            Type.Transitivity transitivity = 1;
+            optional ValueType value_type = 2;
```
flyingsilverfin, line 932: here too, and in `GetOwners`. Anywhere it'ss a restriction? I wonder if that's all the time so it becomes redundant.

proto/user.proto
```diff
@@ -73,7 +73,11 @@ message ClusterUserManager {
     }
 }
 
-message ClusterUser {
+message User {
+    string username = 1;
+    oneof password_expiry {
+        int64 password_expiry_seconds = 2;
+    }
```
flyingsilverfin, line 80: optional?

proto/answer.proto
```diff
@@ -51,7 +51,7 @@ message ConceptMapGroup {
 
 message Numeric {
     oneof value {
-        int64 long_value = 1;
+        sint64 long_value = 1;
```
flyingsilverfin, line 54: This is worth a comment to describe why we have the change, for future reference.

proto/BUILD
```diff
@@ -19,6 +19,33 @@ package(default_visibility = ["//visibility:public"])
...
+
+proto_library(
+    name = "user-proto",
+    srcs = [":user.proto"],
+)
+
+proto_library(
+    name = "database-proto",
+    srcs = [":database.proto"],
+)
```
flyingsilverfin, line 47: if we're being super pedantic I'd put `database` above `user`  ;)

flyingsilverfin (changes requested):
Very good! Some changes/discussions

proto/concept.proto
dmitrii-ubskii: It's always a restriction, outside of `PutAttributeType`. I think tacking on `_filter` would be redundant.

proto/user.proto
dmitrii-ubskii: Nice catch!

proto/concept.proto
```diff
@@ -0,0 +1,953 @@
...
+        }
+        message ResPart {
+            repeated Thing things = 1;
+        }
+    }
+
+    message GetRolePlayers {
+        message Req {}
+        message ResPart {
+            repeated RolePlayer role_types_with_players = 1;
```
flyingsilverfin, line 303: `role_players`

proto/concept.proto
```diff
@@ -0,0 +1,953 @@
...
+            sint64 long = 3;
+            double double = 4;
+            // time since epoch in milliseconds
+            sint64 date_time = 5;
+        }
+    }
+
+    message GetOwners {
+        message Req {
+            optional ThingType filter = 1;
```
flyingsilverfin, line 339: let's call this `thing_type`

proto/concept.proto
```diff
@@ -0,0 +1,953 @@
...
+    message ResPart {
+        oneof res {
+            ThingType.ResPart thing_type_res_part = 1;
+            RoleType.ResPart role_type_res_part = 2;
+        }
+    }
+
+    enum Transitivity {
+        TRANSITIVE = 0;
+        EXPLICIT = 1;
```
flyingsilverfin, line 371: do enums in Prototbuf have a capitalisation convention?

flyingsilverfin (changes requested):

flyingsilverfin (approved):

PR #185, opened 2023-05-26
Add protocol versioning transmitted during connection open

flyingsilverfin:

We add a new connection opening request, which allows transmitting the protocol version for the server to verify.

* Add new service: `connection`, which takes `Connection.Req` and returns `Connection.Res`
* Add new protocol message: `connection.proto`, which contains `Open.Req|Res`
* Add new protocol message: `version.proto`, which holds an Enum encoding the protocol version number


flyingsilverfin:
Merging after review with Ganesh

PR #187, opened 2023-09-15
Unify TypeDB network API & simplify Concept requests

dmitrii-ubskii:

We merge core and cluster gRPC service definitions in order to unify the network interface between the two. As part of this change, we simplify Concept API by merging related requests.

### Build

* migrate to proto3, which allows the use of `optional` in protocol definitions;
* remove nodejs and python protobuf rules — gRPC API will be provided by Rust bindings;

### gRPC Services

* merge core and cluster services and structures into one to unify the API;

### Requests

* merge concept requests where possible by adding arguments (e.g. `GetOwns`, `GetOwnsExplicit`, and `GetOwnsKey` become `GetOwns` with `transitivity` and `annotations` as parameters);
* introduce the `Transitivity` enum and the `RolePlayer` message as part of request restructure.


proto/version.proto
flyingsilverfin, line 1: We should bump the protocol version here I think!

proto/BUILD
```diff
@@ -19,6 +19,34 @@ package(default_visibility = ["//visibility:public"])
 
 load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 
+proto_library(
+    name = "service-proto",
+    srcs = [":service.proto"],
```
flyingsilverfin, line 24: Let's call this `typedb-service` and file `typedb-service.proto`

.factory/automation.yml
```diff
@@ -57,33 +57,6 @@ build:
         export DEPLOY_MAVEN_USERNAME=$REPO_VATICLE_USERNAME
         export DEPLOY_MAVEN_PASSWORD=$REPO_VATICLE_PASSWORD
         bazel run --define version=$(git rev-parse HEAD) //grpc/java:deploy-maven -- snapshot
-    deploy-npm-snapshot:
```
flyingsilverfin, line 60: This will have to be restored

WORKSPACE
```diff
@@ -87,6 +110,21 @@ checkstyle_deps()
 load("@vaticle_dependencies//tool/unuseddeps:deps.bzl", unuseddeps_deps = "deps")
 unuseddeps_deps()
 
+####################################
+# Load @com_google_protobuf #
+####################################
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
```
flyingsilverfin, line 118: I think this shouldn't be in the root WORKSPACE directly

grpc/java/BUILD
```diff
@@ -16,47 +16,26 @@
...
-    deps = [
-        "@maven//:com_google_guava_guava",
-        "@maven//:com_google_protobuf_protobuf_java",
-        "@maven//:io_grpc_grpc_core",
-        "@maven//:io_grpc_grpc_protobuf",
-        "@maven//:io_grpc_grpc_stub",
-        "@maven//:io_grpc_grpc_api",
-        "@maven//:javax_annotation_javax_annotation_api", # gRPC needs this in order to compile in Java 11 and Java 14
+    protos = [
+        "//proto:service-proto",
```
flyingsilverfin, line 25: This will be `//proto:typedb-service` and have `typedb-service.proto`

flyingsilverfin (approved):

PR #202, opened 2024-06-12
Fix type script rules after updated dependencies

farost:
We update dependencies to fix `typedb-protocol` build.

## Implementation
We use the updated rules fixed in [dependencies](https://github.com/vaticle/dependencies/pull/547) and also update the code based on the new requirements of the updates `ts` rules.
We also update `maven` dependencies based on other `sync-dependencies` commits to fix the build.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,3 +7,7 @@ try-import ./.bazel-remote-cache.rc
 build --incompatible_strict_action_env --java_language_version=11 --javacopt='--release 11' --enable_runfiles
 run --incompatible_strict_action_env --java_runtime_version=remotejdk_11
 test --incompatible_strict_action_env --test_env=PATH --cache_test_results=no --java_runtime_version=remotejdk_11
+
+build --@aspect_rules_ts//ts:skipLibCheck=always
+fetch --@aspect_rules_ts//ts:skipLibCheck=always
+query --@aspect_rules_ts//ts:skipLibCheck=always
--- a/dependencies/maven/artifacts.snapshot
+++ b/dependencies/maven/artifacts.snapshot
@@ -1,9 +1,9 @@
 @maven//:com_eclipsesource_minimal_json_minimal_json_0_9_5
 @maven//:com_electronwill_night_config_core_3_6_5
 @maven//:com_electronwill_night_config_toml_3_6_5
-@maven//:com_fasterxml_jackson_core_jackson_annotations_2_10_1
-@maven//:com_fasterxml_jackson_core_jackson_core_2_10_1
-@maven//:com_fasterxml_jackson_core_jackson_databind_2_10_1
+@maven//:com_fasterxml_jackson_core_jackson_annotations_2_16_1
+@maven//:com_fasterxml_jackson_core_jackson_core_2_16_1
+@maven//:com_fasterxml_jackson_core_jackson_databind_2_16_1
 @maven//:com_google_android_annotations_4_1_1_4
 @maven//:com_google_api_grpc_proto_google_common_protos_2_9_0
 @maven//:com_google_auth_google_auth_library_credentials_1_6_0
@@ -25,6 +25,7 @@
 @maven//:com_googlecode_java_diff_utils_diffutils_1_3_0
 @maven//:com_squareup_okhttp_okhttp_2_7_5
 @maven//:com_squareup_okio_okio_1_17_5
+@maven//:com_vdurmont_semver4j_3_1_0
 @maven//:commons_codec_commons_codec_1_13
 @maven//:commons_io_commons_io_2_3
 @maven//:commons_logging_commons_logging_1_2
--- a/dependencies/maven/artifacts.snapshot
+++ b/dependencies/maven/artifacts.snapshot
@@ -1,9 +1,9 @@
 @maven//:com_eclipsesource_minimal_json_minimal_json_0_9_5
 @maven//:com_electronwill_night_config_core_3_6_5
 @maven//:com_electronwill_night_config_toml_3_6_5
-@maven//:com_fasterxml_jackson_core_jackson_annotations_2_16_1
-@maven//:com_fasterxml_jackson_core_jackson_core_2_16_1
-@maven//:com_fasterxml_jackson_core_jackson_databind_2_16_1
+@maven//:com_fasterxml_jackson_core_jackson_annotations_2_16_0
+@maven//:com_fasterxml_jackson_core_jackson_core_2_16_0
+@maven//:com_fasterxml_jackson_core_jackson_databind_2_16_0
 @maven//:com_google_android_annotations_4_1_1_4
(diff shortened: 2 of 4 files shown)


.bazelrc
```diff
@@ -7,3 +7,7 @@ try-import ./.bazel-remote-cache.rc
 build --incompatible_strict_action_env --java_language_version=11 --javacopt='--release 11' --enable_runfiles
 run --incompatible_strict_action_env --java_runtime_version=remotejdk_11
 test --incompatible_strict_action_env --test_env=PATH --cache_test_results=no --java_runtime_version=remotejdk_11
+
+build --@aspect_rules_ts//ts:skipLibCheck=always
+fetch --@aspect_rules_ts//ts:skipLibCheck=always
+query --@aspect_rules_ts//ts:skipLibCheck=always
```
farost, line 13: We started receiving these errors with the newer version of the `ts` rule:
```
ERROR: /private/var/tmp/_bazel_spare/d7e7537e56195a31a080f84b389fd0bd/external/aspect_rules_ts/ts/BUILD.bazel:121:8: configurable attribute "skip_lib_check" in @aspect_rules_ts//ts:options doesn't match this configuration: 

######## Required Typecheck Performance Selection ########

TypeScript's type-checking exposes a flag `--skipLibCheck`:
https://www.typescriptlang.org/tsconfig#skipLibCheck

Using this flag saves substantial time during type-checking.
Rather than doing a full check of all d.ts files, TypeScript will only type-check the code you
specifically refer to in your app's source code.
We recommend this for most rules_ts users.

HOWEVER this performance improvement comes at the expense of type-system accuracy. 
For example, two packages could define two copies of the same type in an inconsistent way.
If you publish a library from your repository, your incorrect types may result in errors for your users.

You must choose exactly one of the following flags:

1. To choose the faster performance put this in /.bazelrc:

    # passes an argument `--skipLibCheck` to *every* spawn of tsc
    # Bazel 6.4 or greater: 'common' means 'any command that supports this flag'
    common --@aspect_rules_ts//ts:skipLibCheck=always

    # Between Bazel 6.0 and 6.3, you need all of this, to avoid discarding the analysis cache:
    build --@aspect_rules_ts//ts:skipLibCheck=always
    fetch --@aspect_rules_ts//ts:skipLibCheck=always
    query --@aspect_rules_ts//ts:skipLibCheck=always

    # Before Bazel 6.0, only the 'build' and 'fetch' lines work.

2. To choose more correct typechecks, put this in /.bazelrc:

    # honor the setting of `skipLibCheck` in the tsconfig.json file
    # Bazel 6.4 or greater: 'common' means 'any command that supports this flag'
    common --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig

    # Between Bazel 6.0 and 6.3, you need all of this, to avoid discarding the analysis cache:
    build --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
    fetch --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
    query --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig

    # Before Bazel 6.0, only the 'build' and 'fetch' lines work.

##########################################################
```

Here I use the fix #2, and these lines were added by @alexjpwalker in `platform`.

grpc/nodejs/BUILD
```diff
@@ -63,6 +63,7 @@ ts_project(
         ":node_modules/@types/node",
         ":node_modules/typescript",
     ],
+    transpiler = "tsc",
```
farost, line 66:
```
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_spare/afea01e34d46fe314768544eee28004b/external/aspect_rules_ts/ts/private/ts_project.bzl", line 246, column 21, in _ts_project_impl
                fail(transpiler_selection_required)
Error in fail: 
######## Required Transpiler Selection ########

You must select a transpiler for ts_project rules, which produces the .js outputs.

Please read https://docs.aspect.build/rules/aspect_rules_ts/docs/transpiler

##########################################################
```

Here I use the transpiler that @alexjpwalker used for `platform`, not sure if anything more special is needed. As I understand, it might affect the runtime speed of the `ts` application if the transpiler is incorrect for this specific app, and I don't think that we want to compare them for our rules. Maybe @alexjpwalker could help us verifying it.

dependencies/vaticle/repositories.bzl
```diff
@@ -7,6 +7,6 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "8ed4fd2b3bfcfbeb572322364e4b3fb8b77064e0", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/farost/dependencies", # TODO: Change back to vaticle when the deps PR is merged!
```
farost, line 10: I won't forget =)

.bazelrc
alexjpwalker: You say you're using fix no.2 which seems more rigorous, but the code diff is suggesting you applied fix no.1.

dependencies/maven/artifacts.snapshot
```diff
@@ -1,9 +1,9 @@
 @maven//:com_eclipsesource_minimal_json_minimal_json_0_9_5
 @maven//:com_electronwill_night_config_core_3_6_5
 @maven//:com_electronwill_night_config_toml_3_6_5
-@maven//:com_fasterxml_jackson_core_jackson_annotations_2_10_1
-@maven//:com_fasterxml_jackson_core_jackson_core_2_10_1
-@maven//:com_fasterxml_jackson_core_jackson_databind_2_10_1
+@maven//:com_fasterxml_jackson_core_jackson_annotations_2_16_0
+@maven//:com_fasterxml_jackson_core_jackson_core_2_16_0
+@maven//:com_fasterxml_jackson_core_jackson_databind_2_16_0
```
alexjpwalker, line 6: Why are these updated?

grpc/nodejs/BUILD
alexjpwalker: Yes, as you have done, I'd rather use `tsc` here for simplicity. In my experience, TypeScript compilation usually completes in under 1 second, so performance is irrelevant.

alexjpwalker (changes requested):

.bazelrc
farost: @alexjpwalker Yeah, I mean fix no.1, the one that I saw for `platform`.

dependencies/maven/artifacts.snapshot
farost: Because it was updated by `sync-dependencies` and I fixed the whole build in this PR. I've updated the description.

alexjpwalker (approved):

dmitrii-ubskii (approved):
LGTM

PR #214, opened 2024-11-21
User management protocol

lolski:
Declare user management messages and services, which lets the caller make user creation, update, deletion, and retrieval call through gRPC.

## Implementation
Declare the associated gRPC services and Protobuf messages for user creation, update, deletion, and retrieval.

### DIFF
--- a/proto/typedb-service.proto
+++ b/proto/typedb-service.proto
@@ -17,20 +17,18 @@ service TypeDB {
     // Connection API
     rpc connection_open (Connection.Open.Req) returns (Connection.Open.Res);
 
-//    // Server Manager API
+    // Server Manager API
     rpc servers_all (ServerManager.All.Req) returns (ServerManager.All.Res);
-//
-//    // User Manager API
-//    rpc users_contains (UserManager.Contains.Req) returns (UserManager.Contains.Res);
-//    rpc users_create (UserManager.Create.Req) returns (UserManager.Create.Res);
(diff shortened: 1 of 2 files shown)


proto/typedb-service.proto
```diff
@@ -17,20 +17,18 @@ service TypeDB {
...
+
+    // User Manager API
+    rpc users_get (UserManager.Get.Req) returns (UserManager.Get.Res);
+    rpc users_all (UserManager.All.Req) returns (UserManager.All.Res);
+    rpc users_contains (UserManager.Contains.Req) returns (UserManager.Contains.Res);
+    rpc users_create (UserManager.Create.Req) returns (UserManager.Create.Res);
+
+    // User API
+    rpc users_update (User.Update.Req) returns (User.Update.Res);
+    rpc users_delete (User.Delete.Req) returns (User.Delete.Res);
```
farost, line 31: I like this, adds symmetry with databases.

proto/user.proto
```diff
@@ -7,76 +7,62 @@ syntax = "proto3";
...
 
-    message PasswordUpdate {
+    message Update {
         message Req {
-            string username = 1;
-            string password_old = 2;
-            string password_new = 3;
+            string name = 1;
+            string password = 2;
+            User user = 3;
```
farost, line 55: What do we use the `User` for in the request? Do we find the existing user by this message, and are `name` and `password` the updated fields then?

farost (approved):
Looking good, left some clarifications comments

PR #216, opened 2025-03-12
Add sign-in request for authentication token retrieval

farost:
A new mechanism of authentication tokens has been introduced to replace the old way of sending usernames and passwords through the network with every request.

Instead, all user credentials (currently, it's usernames and passwords) are sent only:
* as a part of `connection_open` request for authentication and authorization, with a temporary token returned;
* as a part of `sign_in` request for sign ins within an established connection (to change the user or to get a new authentication token).
Then, all further requests are expected to be authenticated only by temporary, less sensitive tokens. 

The approach is extensible to other credential types that can be introduced in the future.

### DIFF
--- a/proto/authentication.proto
+++ b/proto/authentication.proto
@@ -1,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+syntax = "proto3";
+
+package typedb.protocol;
+
+message Authentication {
+
+  // TODO: How to extend it if other credentials appear? Make SignInPassword + SignInXXX, or make username + password optional?
+  message SignIn {
+    message Req {
+      string username = 1;
+      string password = 2;
+    }
+
+    message Res {
+      string token = 1;
+    }
+  }
+}
--- a/proto/authentication.proto
+++ b/proto/authentication.proto
@@ -8,11 +8,14 @@ package typedb.protocol;
 
 message Authentication {
 
-  // TODO: How to extend it if other credentials appear? Make SignInPassword + SignInXXX, or make username + password optional?
   message SignIn {
     message Req {
-      string username = 1;
-      string password = 2;
+      message Password {
+        string username = 1;
+        string password = 2;
+      }
+
+      optional Password password = 1;
     }
 
     message Res {
--- a/proto/authentication.proto
+++ b/proto/authentication.proto
@@ -15,7 +15,10 @@ message Authentication {
         string password = 2;
       }
 
-      optional Password password = 1;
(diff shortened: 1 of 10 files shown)


proto/authentication.proto
```diff
@@ -0,0 +1,28 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+syntax = "proto3";
+
+package typedb.protocol;
+
+message Authentication {
+
+  message SignIn {
```
flyingsilverfin, line 11: Maybe this is `GetToken`? Because you're sending username/pass -> token? Not specifically to do with signing in right here?

flyingsilverfin: Or just `Token`, then you have `Req` and `Res` subtypes!

farost: It's quite a sign in: you sign in with credentials, and you get your access token for further authentication/authorization. Just like in REST, you sign in into a website, and then your token is cached in your cookies, so it is used for authentication/authorization for further requests. I don't mind renaming it to GetToken here though, I'd still name it a "sign in" in HTTP, so I decided to add some symmetry.

farost: Definitely not `Token` as we can `Refresh` it in the future.

flyingsilverfin: Well Token.Refresh makes sense too right?

flyingsilverfin: what does refresh do differently from the regular one?
otherwise we can have `Token.Create.Req/Res` and `Token.Refresh.Req/Res`

farost: Oh, okay, this is a good option. I thought you wanted `Token.Req` and `Token.Res`, and then append `Token.Refresh` to it. 
But I'd still want it to be a part of `Authentication`, because it's the general scope, and we could have some other entities of authentication in the future.

VERSION
```diff
@@ -1 +1 @@
-3.0.0
+3.1.0
```
flyingsilverfin, line 1: needs rebase

proto/version.proto
```diff
@@ -7,7 +7,7 @@ syntax = "proto3";
 package typedb.protocol;
 
 enum Version {
-  reserved 1, 2, 3; // add past version numbers into the reserved range
+  reserved 1, 2, 3, 4; // add past version numbers into the reserved range
   UNSPECIFIED = 0;
-  VERSION = 4;
+  VERSION = 5;
```
flyingsilverfin, line 12: good, thanks!

flyingsilverfin (commented):
Let me know when you're ready to merge and i'll take one more look at it!

VERSION
farost: Thanks!

flyingsilverfin (approved):

PR #221, opened 2025-04-22
Rework transaction options

farost:
Reintroduce transaction options available in TypeDB 3.x. Introduce query options for query-specific configuration.

Remove unused Logic-related messages.

## Implementation
Replace the general `Options` message with `Options.Transaction` and `Options.Query`. Both options are mandatory, but their fields are optional, so the parsing of these messages is straightforward, similar to the 2.x protocol.

Only the options declared in the server's code are introduced.


proto/options.proto
```diff
@@ -7,14 +7,16 @@ syntax = "proto3";
 package typedb.protocol;
 
 message Options {
-//    optional bool infer = 1;
-//    optional bool trace_inference = 2;
-//    optional bool explain = 3;
-    optional bool parallel = 4;
-    optional uint64 prefetch_size = 5;
```
farost, line 14: Removed it as it's not a part of the HTTP protocol and is not used.
I'd also remove `parallel`, but we "declare" it everywhere, so... Let it be. I guess.

flyingsilverfin: Ideally query options, but if not easy it can be a txn option

proto/options.proto
```diff
@@ -7,14 +7,13 @@ syntax = "proto3";
...
-    optional uint64 schema_lock_acquire_timeout_millis = 9;
-    optional bool read_any_replica = 10;
+    message Transaction {
+        optional bool parallel = 1;
+        optional uint64 transaction_timeout_millis = 2;
+        optional uint64 schema_lock_acquire_timeout_millis = 3;
+    }
+
+    message Query {
+        optional bool include_instance_types = 1;
```
flyingsilverfin, line 17: Looks good to me!

proto/options.proto
```diff
@@ -7,14 +7,14 @@ syntax = "proto3";
...
+        optional uint64 prefetch_size = 2;
```
farost, line 18: For simplicity, I left it `uint64`, `u64` in the Rust driver, and `u64` casted to `usize` (as it's used in byte-based channel sizes) in the Rust server

PR #224, opened 2025-05-08
Introduce database import and export protocol messages

farost:
Add `migration` protocol messages for usage in database import and database export operations:
* a unidirectional `database_export` stream from a TypeDB server to export a specific database, similar to TypeDB 2.x;
* a bidirectional `databases_import` stream between a client and a server to import an exported 2.x/3.x TypeDB database into a TypeDB 3.x server from a client.

The format of migration items used for these operations is an extended version of TypeDB 2.x's migration items, so it is backward compatible with 2.x database files. **Important:** it's not intended to import 3.x databases into 2.x servers.

## Implementation

Add `Migration { Item }` message. The format is an extended version of the [2.x protocol](https://github.com/typedb/typedb/blob/2.29.1/migrator/database/data.proto), so it contains "outdated" fields for its compatibility with old databases.

 Add `Migration { Export }` message. This operation consists of a single client `Req { database }` and multiple streamed server responses: 
 
1. An initial response with the schema.
2. An unlimited number of migration items (multiple messages with multiple items in one message for potential optimizations)
3. A `Done` message to signal that the server is ready to close the stream without errors. It can be substituted by a silent stream closure, but I preferred explicitness here.

Add `Migration { Import }` message. This operation consists of a stream of client requests:

1. An initial request with the name of the database and its schema string.
2. An unlimited number of migration items.
3. A `Done` message to signal that the client is finished without errors, and the server can perform the final validation. This `Done` message is required and cannot be removed because the client has to check whether there were finalization errors or not.

 and a stream of server responses (actually, there is either a single `Done` or a single error, but the stream is needed to return errors at any stage of the communication).

### DIFF
--- a/proto/migration.proto
+++ b/proto/migration.proto
@@ -60,6 +60,11 @@ message Migration {
         }
     }
 
+    //               _  _____ _____ _____ _   _ _____ ___ ___  _   _ _
+    //              / \|_   _|_   _| ____| \ | |_   _|_ _/ _ \| \ | | |
+    //             / _ \ | |   | | |  _| |  \| | | |  | | | | |  \| | |
+    //            / ___ \| |   | | | |___| |\  | | |  | | |_| | |\  |_|
+    //           /_/   \_\_|   |_| |_____|_| \_| |_| |___\___/|_| \_(_)
     //
     // ATTENTION: the messages below are used to import multiple versions of TypeDB.
     // DO NOT reorder or delete existing and reserved indices. Be careful while extending this.


proto/database.proto
```diff
@@ -37,10 +41,21 @@ message DatabaseManager {
         message Req {
             string name = 1;
         }
+
         message Res {
             DatabaseReplicas database = 1;
         }
     }
+
+    message Import {
```
farost, line 50: I've placed all migration-related messages in a single file, so it's "encapsulated". However, this introduces an additional layer of optional `client` and `server` in Rust while unpacking this message, so it's a little irritating. Not sure if it's worth it or not. I like this design more, I guess.

proto/error.proto
```diff
@@ -0,0 +1,15 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+syntax = "proto3";
+
+package typedb.protocol;
+
+// This is an emulation of the google ErrorDetails message. Generally, ErrorDetails are submitted via the GRPC error
+// mechanism, but a manual error sending is required in streams
+message Error {
```
farost, line 11: I was planning to reuse this error, but I understood that I don't need non-terminal errors in my protocol. However, I think that it's reasonable to generalize error messages for the whole TypeDB protocol.

proto/version.proto
```diff
@@ -7,7 +7,7 @@ syntax = "proto3";
 package typedb.protocol;
 
 enum Version {
-  reserved 1, 2, 3, 4, 5; // add past version numbers into the reserved range
+  reserved 1, 2, 3, 4, 5, 6; // add past version numbers into the reserved range
   UNSPECIFIED = 0;
-  VERSION = 6;
+  VERSION = 7;
```
farost, line 12: @flyingsilverfin We need to discuss the new versioning approach. I may roll it back.

dmitrii-ubskii (approved):
Looks good, couple notes.

proto/migration.proto
```diff
@@ -0,0 +1,139 @@
...
+            string original_database = 2;
+        }
+
+        message Checksums {
+            int64 entity_count = 1;
+            int64 attribute_count = 2;
+            int64 relation_count = 3;
+            int64 role_count = 4;
+            int64 ownership_count = 5;
+            // 6 was deleted and cannot be used until a breaking change occurs
```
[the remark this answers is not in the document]
farost: This was copied from the 2.x implementation, but I didn't even know there was such a feature. Cool, done!

proto/migration.proto
```diff
@@ -0,0 +1,139 @@
...
+            Done done = 1;
+
+            message Done { }
+        }
+    }
+
+    //
+    // ATTENTION: the messages below are used to import multiple versions of TypeDB.
+    // DO NOT reorder or delete existing and reserved indices. Be careful while extending this.
+    //
```
flyingsilverfin, line 66: bigger please!! like we had it in server - this is very dangerous

farost: ASCII art here we go???

```
//               _  _____ _____ _____ _   _ _____ ___ ___  _   _ _
//              / \|_   _|_   _| ____| \ | |_   _|_ _/ _ \| \ | | |
//             / _ \ | |   | | |  _| |  \| | | |  | | | | |  \| | |
//            / ___ \| |   | | | |___| |\  | | |  | | |_| | |\  |_|
//           /_/   \_\_|   |_| |_____|_| \_| |_| |___\___/|_| \_(_)
```

flyingsilverfin: let's do it :D

flyingsilverfin (approved):
LGTM

PR #230, opened 2025-08-07
Split protocol into 3 separate services

farost:
Split the protocol into 3 services:
* TypeDBService (contains common TypeDB logic): server, database, user, and transaction management, etc.
* TypeDBClusteringService (contains TypeDB logic regarding clustering): replica registration, server status. 
* RaftService (contains Raft logic): replication operations.

## Implementation
Split the old RaftPeeringService into TypeDBClusteringService and RaftService. 

Move separate services and their implementation files into separate subpackages. 

Rename methods. Introduce `ServersGet` method for single server status retrieval. Introduce `registration` method to verify the replica registration operation before submitting it to Raft.


proto/raft_service/replication.proto
```diff
@@ -6,10 +6,7 @@ syntax = "proto3";
 
 package typedb.protocol;
 
-import "proto/server.proto";
-
-// TODO: Rename to Replication
-message Peering {
+message Replication {
```
lolski, line 9: Now that I'm looking into this as a reader, I think the message should be called `Raft` rather than `Replication`. Similarly, `RaftService` should have one rpc endpoint, `raft`.

It's because the message that we exchange are not just replication message, but any messages as defined by the Raft protocol. They can either be "request vote", "append entries" (ie., the message responsible for replication), and "cluster conf change" (ie., add / remove node).

farost: Hmmm okay. I thought that we considered extending this protocol and adding more methods. Your point about vote requests, etc. is reasonable, I'll rename it now.

proto/typedb_clustering_service/typedb_clustering.proto
```diff
@@ -0,0 +1,25 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+syntax = "proto3";
+
+package typedb.protocol;
+
+import "proto/typedb_service/server.proto";
+
+message Registration {
```
lolski, line 11: When does this message gets sent again, and by whom to whom? The reason I'm asking is because now I feel the name `Registration` might not be entirely describe what the message does.

lolski: How does the "registration" relate to the "add node conf change" raft message? In terms of terminology, they're quite synonymous even though the latter is where the bulk of the logic for adding a server takes place.

PR #239, opened 2025-11-19
Introduce clustered servers

farost:

Introduce cluster support in the TypeDB protocol.

In TypeDB 2.x, each database response carried its own replica list (`DatabaseReplicas`). In TypeDB 3.x, servers are the unit of replication, and all databases on that server share the replication context.

- Replace per-database replica discovery with server-level discovery. On connection open, return `ServerManager.All.Res` (servers with replica status) instead of `DatabaseManager.All.Res` (databases with replicas).
- Add `servers_get` RPC to retrieve the specific server's replica status.
- Add `server_version` RPC to retrieve server distribution and version.
- Remove `DatabaseReplicas` message. Instead, return `Database` (name only) instead of `DatabaseReplicas` (name + replica info). Replica topology is a server-level concern.
- Replica information is now on `Server.ReplicaStatus` (replica ID, role, term).

**Breaking**: not compatible with previous versions.

## Implementation

**`server.proto`** — Extended the `Server` message with:
- `optional ReplicaStatus replica_status` (field 2) containing replica ID, role enum, and term
- Nested `Version` request/response messages for the new `server_version` RPC
- New `ServerManager.Get` request/response messages for the new `servers_get` RPC

**`database.proto`** — Simplified:
- Removed `DatabaseReplicas` message and its nested `Replica` message
- Added `name` field (field 1) to the `Database` message (previously it only contained nested schema/delete messages)
- Updated `DatabaseManager.Get/All/Create` responses to return `Database` instead of `DatabaseReplicas`

**`connection.proto`** — Updated handshake:
- `Connection.Open.Res` field 3 changed from `DatabaseManager.All.Res databases_all` to `ServerManager.All.Res servers_all`
- Added `import "proto/server.proto"` dependency

**`typedb-service.proto` → `typedb_service.proto`** — Renamed to follow proto naming conventions (snake_case). Added `servers_get` and `server_version` RPC methods.

For the cleanup purposes within a major update, the fields are reorganized, so the users of this protocol will not be compatible with the older versions.


.factory/automation.yml
```diff
@@ -26,7 +26,7 @@ build:
     deploy-crate-snapshot:
       filter:
         owner: typedb
-        branch: [master]
+        branch: [master, cluster-support-feature-branch]
```
farost: TODO: remove

proto/BUILD
```diff
@@ -102,52 +118,19 @@ proto_library(
...
+    name = "user-proto",
+    srcs = ["user.proto"],
 )
 
 proto_library(
     name = "version-proto",
     srcs = ["version.proto"],
 )
 
-# TODO: THIS SHOULD BE MADE TO STOP EXISTING
```
farost, line 133: Only happy to help

proto/connection.proto
```diff
@@ -6,35 +6,30 @@ syntax = "proto3";
...
     }
 
     message Res {
       uint64 server_duration_millis = 1;
       ConnectionID connection_id = 2;
-
-      // pre-send all databases and replica info
-      DatabaseManager.All.Res databases_all = 3;
-
+      ServerManager.All.Res servers_all = 3;
```
farost, line 27: databases are not actually used or useful. Since replication moved from the database to the server layer, it's only logical to send the list of available servers.

proto/typedb_service.proto
```diff
@@ -24,6 +24,10 @@ service TypeDB {
 
     // Server Manager API
     rpc servers_all (ServerManager.All.Req) returns (ServerManager.All.Res);
+    rpc servers_get (ServerManager.Get.Req) returns (ServerManager.Get.Res);
+
+    // Server API
+    rpc server_version (Server.Version.Req) returns (Server.Version.Res);
```
farost, line 30: Mirroring HTTP

.factory/automation.yml
farost: TODO: Update protocol version since we break backwards compatibility

proto/typedb_service.proto
farost, line 1: Renamed to follow proto naming conventions (snake_case). Can be rollbacked

proto/BUILD
```diff
@@ -8,40 +8,36 @@ load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 
 proto_library(
     name = "typedb-service",
-    srcs = [":typedb-service.proto"],
+    srcs = ["typedb_service.proto"],
```
farost, line 11: https://protobuf.dev/programming-guides/style/ 
While we break some stuff, I thought it might be a good moment to touch this

RELEASE_TEMPLATE.md
```diff
@@ -1,5 +1,8 @@
 Documentation: https://typedb.com/docs/drivers/
 
+**This is an alpha release for CLUSTERED TypeDB 3.x. Do not use this as a stable version of TypeDB.**
```
farost: TODO: Remove

proto/server.proto
```diff
@@ -7,15 +7,42 @@ syntax = "proto3";
...
+        message Req {}
+        message Res {
+            Server server = 1;
+        }
+    }
 }
 
 message Server {
-    string address = 1;
+    optional string address = 1;
```
farost, line 26: It contains ONLY the connection address of the grpc endpoint. No serving addresses, no http addresses. I think it's correct.

grpc/java/BUILD
```diff
@@ -11,21 +12,21 @@ load("@rules_proto_grpc_java//:defs.bzl", "java_grpc_library")
 java_grpc_library(
     name = "typedb-protocol",
     protos = [
-        "//proto:typedb-service",
-        "//proto:server-proto",
-        "//proto:user-proto",
-        "//proto:database-proto",
-        "//proto:migration-proto",
-        "//proto:error-proto",
+        "//proto:analyze-proto",
```
flyingsilverfin: errr do we even still need to support java typedb-protocol

grpc/nodejs/BUILD
```diff
@@ -25,28 +29,30 @@ protoc_gen_ts_bin.protoc_gen_ts_binary(
 ts_grpc_compile(
     name = "typedb-protocol-src",
     deps = [
-        "//proto:typedb-service",
-        "//proto:server-proto",
-        "//proto:user-proto",
-        "//proto:database-proto",
-        "//proto:migration-proto",
-        "//proto:error-proto",
+        "//proto:analyze-proto",
```
flyingsilverfin: and also here? Could delete IMO, less maintenance long term

proto/typedb_service.proto
flyingsilverfin: i like

flyingsilverfin (approved):

grpc/java/BUILD
farost: We don't use it. I just keep supporting it since it's there, maybe somebody depends on it -- it's very cheap for us (you do these global searches of deps anyway). Can be removed easily

grpc/nodejs/BUILD
farost: I'd personally leave this since we still have the `nodejs` dir in `typedb-driver`, which is just not updated. If we decide "nah, we're not doing this in the coming years", then we just need to completely remove both.

grpc/java/BUILD
flyingsilverfin: nah just deleted it

grpc/nodejs/BUILD
flyingsilverfin: i think we can delete it - http seems to work quite well for most people in the js environment IMO. We can always resurrect it if we need to, what do you think @lolski

lolski: Agreed @flyingsilverfin. Less maintenance that way!

PR #246, opened 2026-04-01
Remove replica registration methods from the public proto

farost:
Remove replica registration/deregistration methods from the public protocol as a part of the move to the admin tool. These functions will be available only for the system administrators with access to the server's localhost.

Additionally, remove the unnecessary address fields from `Server` messages.

### DIFF
--- a/proto/server.proto
+++ b/proto/server.proto
@@ -20,21 +20,6 @@ message ServerManager {
             Server server = 1;
         }
     }
-
-    message Register {
-        message Req {
-          uint64 replica_id = 1;
-          string address = 2;
-        }
-        message Res {}
-    }
-
-    message Deregister {
-        message Req {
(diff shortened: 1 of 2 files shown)


proto/server.proto
```diff
@@ -20,27 +20,11 @@ message ServerManager {
...
-        }
-        message Res {}
-    }
 }
 
 message Server {
-    optional string serving_address = 1;
-    optional string connection_address = 2;
-    optional ReplicaStatus replica_status = 3;
+    optional string address = 1;
```
farost, line 26: Two options:
1. We either care about all addresses and all details (so we also share http connection/serving addresses)
2. Or we just require the address needed for gRPC client connections (thus, effective connection address). 

I just cut angles.

PR #251, opened 2026-05-20
Update protocol for introducing 'given' stage

krishnangovindraj:
Updates the protobuf definition for the new 'given' stage.

## Implementation
Adds `given_rows` field to the query request, and the given stage field to the analyzed query.

### DIFF
--- a/proto/query.proto
+++ b/proto/query.proto
@@ -36,10 +36,12 @@ message Query {
         message QueryInputEntry {
             oneof entry {
                 QueryInputEmpty empty = 1;
-                Entity entity = 2;
-                Relation relation = 3;
-                Attribute attribute = 4;
-                Value value = 5;
+                string unparsed_value = 2;
+                // Value value = 3;
+                // Entity entity = 4;
+                // Relation relation = 5;
+                // Attribute attribute = 6;
+
             }
         }
         message QueryInputEmpty {}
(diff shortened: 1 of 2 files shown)


proto/query.proto
```diff
@@ -22,6 +23,24 @@ message Query {
...
+        }
+
+        message GivenEntry {
+            oneof entry {
+                GivenEntryEmpty empty = 1; // Optional None
+                Value value = 2;
+                Thing thing = 3;
+            }
+        }
+        message GivenEntryEmpty {}
```
flyingsilverfin: should be nested inside as just `EmptyEntry` i think

proto/query.proto
```diff
@@ -22,6 +23,25 @@ message Query {
...
+        message GivenRows {
+            repeated string variables = 1;
+            repeated GivenRow rows = 2;
+        }
+
+        message GivenRow {
+            repeated GivenEntry entries = 1;
+        }
+
+        message GivenEntry {
```
flyingsilverfin, line 37: Why don't we reuse the `RowEntry` thing we have from before? That has lists sure, but in the future we'll probably allow entering lists no?

flyingsilverfin: Ah I see we're excluding `Type` ok perhaps this is fine

flyingsilverfin (approved):

